PR #25, opened 2025-01-03
release v2.4.1

samzong:
In accordance with the HAMi release note, update the latest version.


samzong:
@wawa0210 @archlitchi The latest version has supported it.

PR #79, opened 2025-07-15
Improve text userguide/Cambricon-device/ pages

windsonsea:
- remove unecessary headings, that are repeated with that in front matter
- revise text and improve layout


archlitchi (approved):
/lgtm

PR #87, opened 2025-07-18
updated usage example of hygon assigning tasks to specific DCUs

peachest:
With this [PullRequest](https://github.com/Project-HAMi/dcu-vgpu-device-plugin/pull/17) fixed, the id of the hygon DCU should now be unique in the cluster, so you can use `hygon.com/use-uuid` annotation to specify the Pod scheduling range.

Sample documentation for assigning tasks to specific hygon DCUs has been added, taking into account existing documentation

### DIFF
--- a/docs/userguide/Hygon-device/examples/specify-certain-cards.md
+++ b/docs/userguide/Hygon-device/examples/specify-certain-cards.md
@@ -1,0 +1,24 @@
+---
+title: Assign task to certain DCU cards
+---
+
+## Assign task to certain DCU cards
+
+To assign a task to certain DCUs, you need only to assign the `hygon.com/use-gpuuuid` in annotations field.
+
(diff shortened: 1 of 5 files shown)


Nimbus318:
@peachest I forgot to mention earlier — please also update the sidebar config accordingly to include the two new files you added.
https://github.com/Project-HAMi/website/blob/702238bde38f7a6829ef91c8d0efdba38d4c4a1a/sidebars.js#L96-L112

You’ll need to add:
- "userguide/Hygon-device/specify-device-uuid-to-use" at the top level of the items array (alongside specify-device-memory-usage, etc.)
- "userguide/Hygon-device/examples/specify-certain-cards" inside the "Examples" subcategory.

This ensures they show up properly in the documentation sidebar. Thanks again!

PR #89, opened 2025-07-21
Clean up userguide/Metax-device/ pages

windsonsea:
Improve the content by:

- removing unecessary headings if `title` exists
- revising text while keeping same meaning
- fixing some grammar issues
- adding code symbol after backticks


archlitchi (approved):
/lgtm

PR #92, opened 2025-07-25
[zh] sync userguide/Metax-device/ pages

windsonsea:
Update zh text in:

```
i18n/zh/docusaurus-plugin-content-docs/current/userguide/Metax-device/
```


archlitchi (approved):
/lgtm

PR #97, opened 2025-08-05
fix(hygon): corrected a number of documentation errors

peachest:
1. Replace the Cambricon document that was incorrectly placed under the hygon category
2. Translate English comments in code blocks in the Chinese documents into Chinese
3. Add missing yaml language specifiers to the code block
4. Add % postfix to the pod contaienr device cores resource request


windsonsea (commented):
Thanks
/lgtm

Issue #94, opened 2025-07-31
For resources folder, can we make resource folder under i18n as a link to resource folder?


SamYuan1990:
to reduce our images... and assets?

windsonsea:
Could you provide any example to manage those images and assets?

SamYuan1990:
> Could you provide any example to manage those images and assets?

for example
https://github.com/Project-HAMi/website/blob/master/docs/resources/HAMI-VGPU-mind-map-English.png
and
https://github.com/Project-HAMi/website/blob/master/i18n/zh/docusaurus-plugin-content-docs/current/resources/HAMI-VGPU-mind-map-English.png

cc: @archlitchi

windsonsea:
Seems vLLM uses the assets: https://github.com/vllm-project/vllm/tree/main/docs/assets

We need to know how to implement it for both en and i18n.

SamYuan1990:
> Seems vLLM uses the assets: https://github.com/vllm-project/vllm/tree/main/docs/assets
> 
> We need to know how to implement it for both en and i18n.

vLLM uses mkdocs for document service(https://github.com/vllm-project/vllm/blob/main/mkdocs.yaml),  for this repo we using Docusaurus. I don't know the historical reason why we are using Docusaurus.

I traced the code, it seems due to react module for change log and commit history...
I am not sure if https://github.com/sustainable-computing-io/kepler-doc CNCF project kepler can be some thing for maintainers to reference with?

archlitchi:
> > Seems vLLM uses the assets: https://github.com/vllm-project/vllm/tree/main/docs/assets
> > We need to know how to implement it for both en and i18n.
> 
> vLLM uses mkdocs for document service(https://github.com/vllm-project/vllm/blob/main/mkdocs.yaml), for this repo we using Docusaurus. I don't know the historical reason why we are using Docusaurus.
> 
> I traced the code, it seems due to react module for change log and commit history... I am not sure if https://github.com/sustainable-computing-io/kepler-doc CNCF project kepler can be some thing for maintainers to reference with?

The reason is the template is using Docusaurus, we can switch to other solutions:)

mesutoezdil:
Hi @SamYuan1990, @windsonsea, @archlitchi 

Taking a look at the current repo state, I think this problem has effectively been resolved already.

**What was the original problem?**

Duplicate image files living in two places:
- `docs/resources/HAMI-VGPU-mind-map-English.png`
[...]

PR #101, opened 2025-08-22
example with mkdocs

SamYuan1990:
as discussed on today's community session
here is a local preview for mkdocs version of HAMi document.
feature list
```
- index page
Nav
- Documentation
    core conceptes
        what's
        arch
- Blog
- Changelog

- search
- i18n for zh and en
```

It has to be a breaking change, so with this PR, just have a version of us to discuss.
for maintainers, please run locally to try with it.

related with #94


windsonsea:
If the need is only for handling image resources, adopting a new website builder isn’t really necessary.
Major breaking changes usually introduce more issues and bugs.

rootsongjc:
Thanks for putting this together and sharing the preview! @SamYuan1990 

Just a quick update from my side — I recently finished a major refresh of the whole website. We decided to keep using **Docusaurus** and have already migrated and reorganized the content there.

Because of that, this MkDocs version probably doesn’t make much sense for us to move forward with anymore. Really appreciate the effort and the idea though — it was useful for discussion when we were exploring options.

Let’s keep the current direction on Docusaurus for now.

PR #103, opened 2025-09-01
docs: simplify GPU runtime configuration using nvidia-ctk & fix k8s pod container crashes 

IsQiao:
Context:
When I use the original document, k8s's pod is only valid for the first time, and after restarting, the pod is very fragile, and there is a small probability that it will be successful.

Changes:
- Replace manual Docker daemon.json configuration with nvidia-ctk runtime configure --runtime=docker
- Replace manual containerd config.toml configuration with nvidia-ctk runtime configure --runtime=containerd
- Update both English and Chinese documentation for easier setup process

Reference:
- https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html
- https://www.reddit.com/r/kubernetes/comments/yvj4s4/setting_containerd_default_runtime_to_nvidia/
- https://www.cnblogs.com/handsomer/p/19068098

### DIFF
--- a/docs/installation/prequisities.md
+++ b/docs/installation/prequisities.md
@@ -31,18 +31,10 @@ sudo apt-get update && sudo apt-get install -y nvidia-container-toolkit
 
 #### Configure Docker
 
-When running Kubernetes with Docker, edit the configuration file, typically located at `/etc/docker/daemon.json`, to set up `nvidia-container-runtime` as the default low-level runtime:
+When running Kubernetes with Docker, use the `nvidia-ctk` tool to automatically configure Docker:
 
-```json
-{
-    "default-runtime": "nvidia",
-    "runtimes": {
-        "nvidia": {
-            "path": "/usr/bin/nvidia-container-runtime",
-            "runtimeArgs": []
-        }
-    }
-}
+```bash
+sudo nvidia-ctk runtime configure --runtime=docker
 ```
 
 And then restart Docker:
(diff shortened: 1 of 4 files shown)


docs/installation/prequisities.md
```diff
@@ -31,18 +31,10 @@ sudo apt-get update && sudo apt-get install -y nvidia-container-toolkit
...
-        "nvidia": {
-            "path": "/usr/bin/nvidia-container-runtime",
-            "runtimeArgs": []
-        }
-    }
-}
+When running Kubernetes with Docker, use the `nvidia-ctk` tool to automatically configure Docker:
+
+```bash
+sudo nvidia-ctk runtime configure --runtime=docker
```
wawa0210, line 37: How to install nvidia-ctk

IsQiao: The NVIDIA Container Toolkit already contains the tool

IsQiao: https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html#configuring-docker

archlitchi:
what leads to this 'fragile' issue, i wonder, because even if the pod is restarted, the daemon.json or config.toml should be the same after your configuration.

IsQiao:
> what leads to this 'fragile' issue, i wonder, because even if the pod is restarted, the daemon.json or config.toml should be the same after your configuration.

---
@archlitchi  No, the configuration files are not the same. Also, it is not the pod restart, but all pods of this k8s node have problems because they use the default nvidia-container-runtime.

that is the different(the left side is generated by the tool).
image

archlitchi (approved):
/lgtm

Maybe you could fix the README.md of HAMi as well

PR #156, opened 2026-04-02
[Docs] replace 您 with 你 and remove unnecessary 请

windsonsea:
The current website has both 您 and 你。

This PR standardizes the tone of the documentation by:

- replacing "您" with "你"
- removing unnecessary polite wording such as "请"

As an open source website and doc, it's not very necessary to use many polite wordings like 您、请


archlitchi (approved):
/lgtm

PR #164, opened 2026-04-15
fix(announcement-bar): show bar only with localized copy

rootsongjc:
Load announcement bar messages from locale-specific i18n files (en/zh) using the current Docusaurus locale, and skip rendering when the localized message is missing or empty.

Also clear the English announcement text to disable the expired event banner and prevent stale or unintended announcements from displaying.fix(announcement-bar): show bar only with localized copy

Signed-off-by: Jimmy Song <jimmy@dynamia.ai>


windsonsea (commented):
Thanks
/lgtm
/approve

PR #165, opened 2026-04-17
 fix incorrect hygon DCU verification command

xrwang8:
Corrects the command used to verify virtual DCU allocation inside containers.

  The actual driver command is `hy-smi virtual -show-device-info`, but the documentation incorrectly references `hy-virtual -show-device-info`. This PR updates all relevant docs to use the correct command.


windsonsea (commented):
/approve

PR #172, opened 2026-04-19
fix: update outdated Kubernetes version example from v1.16.8 to v1.29.0

mesutoezdil:
## What

The documentation used v1.16.8 as an example Kubernetes version in two places. v1.16.8 was released in 2019 and reached end-of-life years ago. Any user following this example today would be using a version that is no longer supported.

Updated both files to use v1.29.0 as the example, which is a recent supported release.

## Files changed

docs/installation/online-installation.md (line 27 and 30)
docs/get-started/deploy-with-helm.md (line 113 and 117)


windsonsea (commented):
/lgtm
/approve

PR #232, opened 2026-04-24
docs: 'so to minimize' -> 'so as to minimize' in awsneuron intro

mesutoezdil:
Minor prose fix in the topology awareness bullet. 'so to minimize' reads awkwardly; 'so as to minimize' is standard.


windsonsea (commented):
/lgtm
/approve

PR #234, opened 2026-04-26
docs: 'a HTML' -> 'an HTML' in contribute-docs

mesutoezdil:
Article agreement fix. HTML starts with a vowel sound ("aitch"), so it takes 'an'.


windsonsea (commented):
/lgtm
/approve

PR #284, opened 2026-04-30
[i18n-zh] Sync offline-installation and prerequisites

windsonsea:
Update zh text in:

```
i18n/zh/docusaurus-plugin-content-docs/current/installation/offline-installation.md
i18n/zh/docusaurus-plugin-content-docs/current/installation/prerequisites.md
```


rootsongjc (approved):
/lgtm

PR #308, opened 2026-05-05
blog: remove emoji from meetup beijing and shanghai 2025 posts

mesutoezdil:
Remove the 👉 emoji from the language-switch note in the Beijing and Shanghai meetup 2025 blog posts.


rootsongjc:
This blog is not visible for the website, we don't need to change it. This articles about the HAMi Meetup event is only available in Chinese.

PR #320, opened 2026-05-06
docs: remove filler words from versioned doc snapshots

mesutoezdil:
Remove filler words from historical doc snapshots (v1.3.0 through v2.8.0).


windsonsea (approved):
/lgtm
/approve

PR #369, opened 2026-05-11
fix: make docs/ the default version and sync v2.8.0 snapshot

mesutoezdil:
## Problem

The website defaulted to the v2.8.0 versioned snapshot instead of the current docs. This caused two issues:

1. All improvements merged to `docs/` were invisible on the live site despite being in master.
2. Contributors edited `versioned_docs/version-v2.8.0/` directly to make changes visible, which caused the snapshot to diverge from `docs/` over time.

## Changes

**Commit 1 - Sync 97 files in v2.8.0 snapshot:**
Copies all files where `versioned_docs/version-v2.8.0/` had fallen behind `docs/`. Covers contributor guides, installation, userguide, developers, core-concepts, get-started, faq, troubleshooting, key-features, and releases.

**Commit 2 - Set `lastVersion: 'current'` in docusaurus.config.js:**
Makes `docs/` the default shown at `/docs/`. Versioned snapshots remain accessible at `/docs/v2.8.0/`, `/docs/v2.7.0/`, etc. The version dropdown labels current docs as "latest".

After this change, any PR merged to `docs/` immediately appears on the live site. No more manual syncing of versioned snapshots.

## Test plan

- `npm run build:fast` passes (English only)
- `npm run build` passes (English + Chinese)
- `diff -rq docs/ versioned_docs/version-v2.8.0/ --exclude="*.json"` reports 0 differing files


rootsongjc (approved):
/lgtm

PR #372, opened 2026-05-12
[i18n-zh] Update vastai pages

windsonsea:
Update zh text in:

```
i18n/zh/docusaurus-plugin-content-docs/current/userguide/vastai/enable-vastai-sharing.md
i18n/zh/docusaurus-plugin-content-docs/current/userguide/vastai/examples/default-use.md
```


rootsongjc (approved):
/lgtm

PR #373, opened 2026-05-13
[i18n-zh] Update userguide/configure.md

windsonsea:
Sync zh text with the english upstream:

```
i18n/zh/docusaurus-plugin-content-docs/current/userguide/configure.md
```


mesutoezdil:
@archlitchi @windsonsea the sync looks good overall, but the **Container Configs: Env** section is missing from this PR. The English file has it at the bottom with `GPU_CORE_UTILIZATION_POLICY` and `CUDA_DISABLE_CONTROL`. Those two entries need to be translated and added to the ZH file as well to complete the sync.

Also, the English source has a small typo: `using \`-set\`` should be `using \`--set\``. Not introduced by this PR, but worth fixing upstream too.

archlitchi (commented):
/lgtm

PR #381, opened 2026-05-14
fix: add language specifiers to 686 bare code fences across all docs

mesutoezdil:
Adds missing language specifiers (yaml, bash, json, console, text) to 686 bare code fences across docs/, versioned_docs/, i18n/, and blog/. Language is inferred from block content. No content changes.


windsonsea (commented):
/lgtm
/approve

PR #401, opened 2026-05-21
fix(i18n): add missing Chinese translations for 29 UI strings in code.json

mesutoezdil:
29 UI strings in `i18n/zh/code.json` had no Chinese translation
they still showed raw English text to Chinese visitors. This PR fills all gaps.


mesutoezdil:
Closing - Chinese translations should not be modified in this PR batch.

PR #441, opened 2026-06-12
feat: add RSS and Atom feed options to blog plugin

### LINKED ISSUE #147: rss feeds for agent to read?
ref https://project-hami.io/zh/changelog/rss.xml

do we consider add https://project-hami.io/zh/blog and https://project-hami.io/zh/case-studies into rss feed for AI/Agent to get HAMi community's update?

mesutoezdil:
Adds explicit `feedOptions` to the blog preset in `docusaurus.config.js`.

## What changes

- `type: ['rss', 'atom']` - generates both feed formats
- `title`, `description`, `copyright` - production-ready feed metadata
- Exposes `/blog/rss.xml`, `/blog/atom.xml` for EN
- Exposes `/zh/blog/rss.xml`, `/zh/blog/atom.xml` for ZH

Both feeds verified present in build output.

Closes #147

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -396,6 +396,12 @@ module.exports = {
         blog: {
           showReadingTime: true,
           editUrl: "https://github.com/Project-HAMi/website/tree/master/",
+          feedOptions: {
+            type: ['rss', 'atom'],
+            title: 'HAMi Blog',
+            description: 'Latest news and updates from the HAMi project',
+            copyright: `Copyright ${new Date().getFullYear()} HAMi Authors`,
+          },
         },
         theme: {
           customCss: require.resolve("./src/css/custom.css"),


rootsongjc:
Nice improvement, and thanks for verifying both EN and ZH feeds in the build output.

One i18n point worth flagging. `feedOptions.title` and `description` apply globally, so the generated `/zh/blog/rss.xml` and `/zh/blog/atom.xml` will carry the English metadata:

- `title: 'HAMi Blog'`
- `description: 'Latest news and updates from the HAMi project'`

For a Chinese-audience feed, a localized title and description would read better in a feed reader. Docusaurus doesn't support per-locale `feedOptions` on a single blog instance out of the box, so the realistic options are:

1. Keep shared metadata as it is now. Fine as a baseline.
2. Run two blog plugin instances (one per locale) with localized `feedOptions`. More config, but gives true i18n metadata.

Not a blocker for this PR. Just raising it because ZH subscribers will see the English strings in their feed reader.

mesutoezdil:
Thanks for the review.

Keeping shared metadata as the baseline for now. The English strings in ZH feed readers are acceptable at this stage.

Opened #452 to track ZH-localized feedOptions via a second blog plugin instance.

windsonsea (commented):
/lgtm
/approve

PR #458, opened 2026-06-20
remove duplicate Enflame in v2.9.0 device support list

mesutoezdil:
Enflame appeared twice in the device support list in both the EN and ZH versions of the v2.9.0 release blog.

EN: "...Iluvatar, Enflame, AWS Neuron..." - second Enflame removed.
ZH: "...Iluvatar、Enflame、AWS Neuron..." - Enflame removed (燧原 already present earlier in the list).


windsonsea (commented):
/lgtm
/approve

PR #459, opened 2026-06-20
fix KE Holdings case study URL

mesutoezdil:
`src/data/caseStudies.js` used `/case-studies/ke-holdings/` but the canonical CNCF slug is `/case-studies/ke-holdings-inc/`, which is what both the EN and ZH v2.9.0 blog posts already use.


rootsongjc (approved):
/lgtm

PR #461, opened 2026-06-20
add missing Vastai VA16 row to ZH device support table

mesutoezdil:
The EN device support table has `| GPU | Vastai | VA16 |` (added in v2.9.0) but both the current and v2.9.0 versioned ZH translations were missing this row.

Added `| GPU | 瀚博（Vastai） | VA16 | 是 | 是 | 否 |` to both ZH files.


windsonsea (commented):
/lgtm
/approve

PR #463, opened 2026-06-20
remove false .nvmrc reference from README

mesutoezdil:
README says Node.js is "configured in `.nvmrc`" but no `.nvmrc` file exists in the repo.

Removed the parenthetical.


rootsongjc (approved):
/lgtm

PR #468, opened 2026-06-21
fix: remove stale kernel version requirement from prerequisites

mesutoezdil:
Kernel 3.10 is from 2013 and EOL. No supported Kubernetes version runs on kernel 3.10. This entry signals outdated docs without adding value. Removes it.


windsonsea (commented):
/lgtm
/approve

PR #469, opened 2026-06-21
fix: update minimum Kubernetes version in prerequisites

mesutoezdil:
The stated minimum of k8s 1.18 is outdated. HAMi uses APIs and features that require at least 1.23. Updates the minimum to 1.23.


windsonsea (commented):
/lgtm
/approve

PR #475, opened 2026-06-21
fix: use canonical AWS homepage URL in ecosystem section

mesutoezdil:
`www.aws.com` redirects to `aws.amazon.com` but using the non-canonical URL is fragile. Change to the canonical URL.


windsonsea (commented):
/lgtm
/approve

PR #477, opened 2026-06-21
fix: remove deprecated --short flag from kubectl version command

mesutoezdil:
\`kubectl version --short\` was removed in Kubernetes 1.28. Users on any modern cluster will get an error. Removes the flag so the command works.


windsonsea (commented):
/lgtm
/approve

PR #480, opened 2026-06-21
fix: remove duplicate Enflame and incorrect AMD from v2.9.0 blog device list

mesutoezdil:
The device list in the v2.9.0 blog post listed "Enflame" twice and included "AMD" which is not yet a supported device (it is listed as planned in the roadmap). Removes the duplicate and the incorrect entry.

Note: this PR also removes AMD (issue 15) from the list since both fixes are on the same line.


windsonsea (commented):
/lgtm
/approve

PR #481, opened 2026-06-21
fix: use correct namespace in v2.9.0 upgrade command

mesutoezdil:
The upgrade command in the v2.9.0 blog post used `-n hami-system` but the standard installation uses `-n kube-system` (as documented in online-installation.md). Using the wrong namespace would upgrade the wrong Helm release.


windsonsea (commented):
/lgtm
/approve

PR #483, opened 2026-06-21
fix: use Huawei Ascend as manufacturer name in roadmap

mesutoezdil:
The roadmap used "Ascend" as the manufacturer name but device-supported.md and CNCF materials use "Huawei Ascend". Standardizes to the more specific and consistent name.


windsonsea (commented):
/lgtm
/approve

rootsongjc:
One concern is that this PR only changes the roadmap while other documents still use "Ascend" mostly. We should probably decide whether the project prefers vendor names (NVIDIA, AMD, Huawei Ascend) or accelerator brands (CUDA, ROCm, Ascend, etc.) first, otherwise we may introduce inconsistent naming across the website.

Perhaps this should be discussed project-wide before merging.

mesutoezdil:
Thanks for the feedback @rootsongjc. You are right 
it is done now

docs/installation/how-to-use-volcano-ascend.md
```diff
@@ -1,25 +1,25 @@
...
+- Heterogeneous Huawei Ascend cluster
 
 This feature is only available in volcano >= 1.14.
 
 ## Quick Start
 
 ### Prerequisites
 
-[ascend-docker-runtime](https://gitcode.com/Ascend/mind-cluster/tree/master/component/ascend-docker-runtime)
+[ascend-docker-runtime](https://gitcode.com/Huawei Ascend/mind-cluster/tree/master/component/ascend-docker-runtime)
```
rootsongjc: Be careful not to modify the fields in the link.

mesutoezdil: y r right, done

PR #485, opened 2026-06-21
fix: add Ascend 910C to supported device table

mesutoezdil:
The v2.9.0 release added support for Ascend 910C (HAMi-core mode) but the device-supported table was not updated. Adds 910C to the Huawei Ascend row.


rootsongjc (approved):
Can you update the Chinese and latest version as well?

- versioned_docs/version-v2.9.0/userguide/device-supported.md
- i18n/zh/docusaurus-plugin-content-docs/current/userguide/device-supported.md
- i18n/zh/docusaurus-plugin-content-docs/version-v2.9.0/userguide/device-supported.md

rootsongjc (approved):
/lgtm

PR #486, opened 2026-06-21
fix: update HAMi-DRA version reference to v0.2.0

mesutoezdil:
The gpu-virtualization doc referenced HAMi-DRA v0.1.0 but v0.2.0 shipped with v2.9.0.


windsonsea (commented):
/lgtm
/approve

PR #487, opened 2026-06-21
fix: mark DRA support as completed in roadmap

mesutoezdil:
DRA support shipped as GA in v2.9.0 (HAMi-DRA subproject). The roadmap checkbox is still unchecked. Marks it as complete.


windsonsea (commented):
/lgtm
/approve

PR #489, opened 2026-06-21
fix: improve wording in releases page versioning section

mesutoezdil:
Two wording issues in the same sentence on the releases page:
- "Whenever a release is released" is redundant
- "mirror repository" confuses English readers (sounds like a git mirror; the intended meaning is container registry)

Fixes both in one commit since they are in the same sentence.


windsonsea (commented):
/lgtm
/approve

PR #496, opened 2026-06-25
Update mermaid diagrams for clarity in KAI Scheduler integration

rootsongjc:
Enhance the clarity and consistency of mermaid diagrams related to the KAI Scheduler integration, improving the overall understanding of the workflow and GPU memory isolation process.


mesutoezdil:
Hey Jimmy, a few things though:

- shouldn't the commit prefix be `docs:` or `style:` rather than `fix:`?
- collapsing all 5 schedulers into one node loses the fan-out visual, and KAI's individual highlight kind of made the point of the post - worth keeping?
- `DAEMON` now points to `WEBHOOK` instead of `CONTAINER` - intentional?
- the warning/checkmark emojis in the before/after diagram carried over, can we remove those?

rootsongjc:
I can't modify the name of this branch right now, I will use standardized naming in the future.

PR #510, opened 2026-06-26
fix: add missing -- in kubectl exec in versioned_docs

mesutoezdil:
Add missing -- separator before nvidia-smi in kubectl exec commands across all versioned docs (v1.3.0 through v2.9.0).


windsonsea (commented):
/lgtm
/approve

PR #512, opened 2026-06-26
fix: correct helm.sh zh URL in versioned_docs

mesutoezdil:
Replace helm.sh/zh/docs/ with helm.sh/docs/ across all versioned docs (v1.3.0 through v2.9.0).


windsonsea (commented):
/lgtm
/approve

PR #518, opened 2026-07-02
fix: remove npm as project dependency

mesutoezdil:
npm should not be listed as a dependency in package.json. It is the package manager, not a project package.

## Test plan
- npm run lint:md passes

## Summary by CodeRabbit

* **Chores**
  * Removed an unused package entry from the dependency list, simplifying project maintenance.


windsonsea (commented):
/lgtm
/approve

PR #522, opened 2026-07-02
fix: correct organizationName to match actual GitHub org

mesutoezdil:
organizationName was set to hami-io but the actual GitHub org is Project-HAMi. This only affects npm run deploy, which is unused since the site deploys via Netlify. Fixed it to be correct in case anyone runs it.

## Summary by CodeRabbit

* **Chores**
  * Updated the site’s organization name in the documentation configuration to reflect the new Project-HAMi branding.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -56,7 +56,7 @@ module.exports = {
       onBrokenMarkdownLinks: "throw",
     },
   },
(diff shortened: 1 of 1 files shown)


windsonsea (commented):
/lgtm
/approve

PR #524, opened 2026-07-02
fix: mirror hotlinked blog image into static assets

mesutoezdil:
One blog post (en and zh) pulled an image from a personal fork on GitHub. That link breaks if the fork ever moves or gets deleted. Downloaded the image into static/img/blog/ and pointed both posts at the local copy.

## Summary by CodeRabbit

* **Documentation**
  * Updated a blog image reference to use a local site asset path instead of an external URL.
  * Applied the same image path update to the Chinese localized version of the post.


windsonsea (commented):
/lgtm
/approve

PR #525, opened 2026-07-02
fix: replace broken check-links script with working tool

mesutoezdil:
check-links called npx html-link-checker, a package that does not exist on npm. The script always failed with a 404 for anyone who ran it. Swapped in linkinator, which is real and actively maintained.


windsonsea (commented):
/lgtm
/approve

PR #531, opened 2026-07-03
chore: remove dead zh translation keys

mesutoezdil:
i18n/zh/code.json had two old keys, HAMi is a and sandbox project, from an earlier version of the homepage code. No component uses Translate with these ids anymore, the homepage now builds its bilingual text with plain ternaries. These keys never render, removing them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated Chinese localization content by removing two outdated message entries.

### DIFF
--- a/i18n/zh/code.json
+++ b/i18n/zh/code.json
@@ -251,12 +251,6 @@
   "Supporters": {
     "message": "支持者"
   },
-  "HAMi is a": {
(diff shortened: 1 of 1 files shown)


rootsongjc (approved):
/lgtm

windsonsea (commented):
/lgtm
/approve

PR #538, opened 2026-07-03
docs: add missing pprof example to zh profiling guide

mesutoezdil:
The zh version of the memory profiling section stopped right after the two pprof commands. The full worked example with real pprof output that the english doc has was missing. Added it, command output stays untranslated as-is, only the intro line was translated.

## Summary by CodeRabbit

* **Documentation**
  * Added a new example for allocation profiling in the memory performance analysis section.
  * Included sample interactive profiling output, a summary view, and source-level navigation examples to help readers interpret profiling results.


windsonsea (commented):
/lgtm
/approve

PR #546, opened 2026-07-03
docs: fix deploy instructions to match actual netlify hosting

mesutoezdil:
README.md and AGENTS.md both documented npm run deploy for GitHub Pages as the deployment step, and README listed GitHub Pages as a hosting option. No workflow runs gh-pages and the site only deploys through Netlify. Fixed both files to describe the real deploy path.


windsonsea (commented):
/lgtm
/approve

PR #554, opened 2026-07-05
chore: bump github actions to latest majors in ci

mesutoezdil:
checkout was pinned to v4, setup-node to v4, both several majors behind. Bumped to the latest major (checkout v7, setup-node v6). Checked setup-node v6 docs, explicit cache: npm input still works the same, only auto-detection changed for workflows that omit the cache input.

## Summary by CodeRabbit

* **Chores**
  * Updated the documentation health workflow to use newer versions of core GitHub Actions.
  * No user-facing behavior changed; build, lint, format, and link-check steps remain the same.

### DIFF
--- a/.github/workflows/docs-health.yml
+++ b/.github/workflows/docs-health.yml
@@ -15,10 +15,10 @@ jobs:
     name: Lint, Format, Build & Links
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


windsonsea (commented):
/lgtm
/approve

PR #555, opened 2026-07-05
chore: bump minor and patch npm dependencies

mesutoezdil:
Bumped 8 packages to their latest minor or patch version: fortawesome icon packages, docusaurus-search-local, swc/core, asciinema-player, markdownlint-cli, prettier. Left docusaurus itself alone (separate PR already open for that) and left react, react-dom, clsx alone since those are major version jumps that need their own evaluation.


rootsongjc (approved):
/lgtm

PR #562, opened 2026-07-08
fix(site): restore generated-site link checks

spencercjh:
## Summary
- replace the ineffective `linkinator ./build --skip '^https?://'` path with a real generated-site crawl against a served build
- add `scripts/check-links.sh` as the pure-`linkinator` orchestration entrypoint and wire `check:all` / `docs-health.yml` through it
- make `DOCUSAURUS_SITE_URL` overrideable during link-check builds so canonical/og same-site absolute URLs resolve locally
- repair the currently exposed same-site broken links, including route mismatches and the broken Beijing meetup image path

## Why
The previous CI command looked green but actually scanned `0 links`, so generated-site regressions could slip through. This PR makes the link check exercise the built site for real and fixes the broken routes it exposes on the current branch.

## Notes
- this PR intentionally keeps the current whole-site crawl scope rather than narrowing to docs-only
- `scripts/check-links.sh` does not implement a custom checker; it only starts the static server, derives external-host skip rules from generated HTML, and invokes `linkinator`
- there is no repository PR template under `.github`, so this body is written directly

## Validation
- `npm run check:all`
- `npm run check-links`
- `npm run lint`
- `npm run format:check`
- `gh act workflow_dispatch -W .github/workflows/docs-health.yml -j check -P ubuntu-latest=catthehacker/ubuntu:act-latest --no-skip-checkout`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved documentation and blog links so they resolve correctly within the site and localized pages.
  * Updated site settings to use the correct base URL for generated links and canonical pages.
* **New Features**
  * Added richer tag metadata for blog posts and documentation tutorials, improving navigation and discovery.
* **Chores**
  * Strengthened link validation during site checks for more reliable documentation builds.
  * Updated ignore rules to skip local workspace files.

PR #563, opened 2026-07-08
fix: capitalize MIG in remaining zh aws install pages

mesutoezdil:
PR #562 fixed MIG casing in current and v2.7.0 but missed v2.8.0 and v2.9.0. v2.9.0 is the live default version. Same one-line fix applied to both.

## Summary by CodeRabbit

* **Documentation**
  * Updated the NVIDIA device example text in the Chinese installation docs to use “MIG” with the correct capitalization.
  * Kept the existing documentation links unchanged.


rootsongjc:
Can you fix the case of other errors at the same time? Not just the word MIG. And this word also appears in many documents.

PR #588, opened 2026-07-12
fix: remove orphaned unlinked pages from versioned docs

mesutoezdil:
Two kinds of leftover pages found across versioned_docs and its zh translations, neither linked from any sidebar but both still live at real URLs:

- 7 zh roadmap.md files (v1.3.0, v2.4.1, v2.5.0, v2.5.1, v2.6.0, v2.7.0, v2.8.0) titled Karmada Roadmap, scaffold content from an unrelated CNCF project, not present in the matching English versions.
- A duplicate troubleshooting-copy directory in versioned_docs/version-v2.6.0, byte-identical to the real troubleshooting directory.

Verified with a full docusaurus build.


rootsongjc (approved):
/lgtm

PR #601, opened 2026-07-12
fix: correct Helm template syntax in dynamic-mig documentation

mesutoezdil:
resourceCountName and related keys in the device-configmap.yaml example used { { .Values.x } } with extra spaces around the braces, which is not valid Helm/Go template syntax (it renders as literal text instead of substituting the value). Changed to {{ .Values.x }} across docs, the current and v2.9.0 zh translations, and the v2.5.1, v2.6.0, v2.7.0, and v2.9.0 versioned copies. Recovered from an old unmerged local branch, the bug was still present in current master.


windsonsea (commented):
/lgtm
/approve

Issue #628, opened 2026-07-22
Fix 38 open npm dependency vulnerabilities reported by LFX Insights


mesutoezdil:
## Problem

LFX Insights reports 38 open vulnerabilities across npm dependencies (2 critical, 19 high, 23 medium, 8 low, median CVSS 6.2). A local `npm audit` on current master confirms 47 findings (2 critical, 11 high, 33 moderate, 1 low).

All of them are in the Docusaurus build toolchain (transitive deps such as `webpack-dev-server`, `undici`, `ws`, `minimatch`, `node-tar`, `fast-uri`, `shell-quote`, `js-yaml`, `serialize-javascript`). The published site is static HTML, so runtime exposure is low, but the findings are fixable and keep the security score red.

## Status

- #555 bumps minor and patch npm dependencies (open since Jul 5)
- #556 applies `npm audit fix` for the tooling vulnerabilities (open since Jul 5)
- New advisories were published Jul 15-21 (`websocket-driver`, `fast-uri`, `svgo`, `linkify-it`, `webpack-dev-server`, `body-parser`, `brace-expansion`), so #556 will be refreshed against latest master to cover them

Note for maintainers: #555 and #556 both touch `package-lock.json`, so whichever merges second needs a trivial rebase.

Ref: https://insights.linuxfoundation.org/project/hami/repository/project-hami_website/security

mesutoezdil:
Done. LFX Insights now reports **0 open vulnerabilities**, and the repository has **0 open Dependabot alerts**.

The work landed across several pull requests: #556, #566, #573, #668, #669 and #673.

The last one left was `brace-expansion` (CVE-2026-14257), reaching us through `@docusaurus/core > serve-handler > minimatch@3`. It needs a note, because it cannot be fixed the obvious way:

- The advisory lists `5.0.8` as the first patched version, but that release is ESM-first and its CommonJS build exports an object. `minimatch@3` calls the module directly as a function, so forcing `5.0.8` throws `expand is not a function` and breaks `npm run serve`. `serve-handler@6.1.7` is the latest release and hard-pins `minimatch@3.1.5`, so there is no upgrade path from our side.
- #673 pins `1.1.17` instead, the upstream backport of the same fix to the 1.x line. Its `index.js` adds the `EXPANSION_MAX_LENGTH` bound and cites CVE-2026-14257 in the code comment.
[...]

PR #556, opened 2026-07-05
chore: npm audit fix for dev tooling vulnerabilities

mesutoezdil:
## What

Runs `npm audit fix` (no `--force`, no semver-major changes) against latest master. Only `package-lock.json` changes.

## Why

LFX Insights reports 38 open dependency vulnerabilities for this repo. This PR resolves 20 of the 47 local `npm audit` findings, including both critical ones (`shell-quote`, `websocket-driver`) and most high-severity ones (`ws`, `undici`, `fast-uri`, `brace-expansion`, `linkify-it`, `svgo`, `minimatch`, `node-tar`).

Refreshed on Jul 22 to cover advisories published Jul 15-21.

The remaining 27 findings (25 moderate, 2 high) are transitive deps of the Docusaurus toolchain (`webpack-dev-server`, `serialize-javascript`, `js-yaml` via `markdownlint-cli`) and need breaking upgrades, tracked in #628.

## Testing

`npm run build:fast` passes after the change.

Part of #628

PR #630, opened 2026-07-22
chore: fix remaining npm audit findings with overrides

mesutoezdil:
## What

Closes the audit findings that plain `npm audit fix` (#556) cannot reach, via npm `overrides` scoped to the parents that need them, pinned to exact versions:

- `markdownlint-cli` -> `js-yaml` 4.3.0 (high, GHSA-52cp-r559-cp3m, GHSA-h67p-54hq-rp68), `markdown-it` 14.3.0 (moderate, GHSA-6v5v-wf23-fmfq) and `run-con` 1.3.2 (1.3.3 pulls `ini@7`, which requires Node 22+ while CI runs Node 20)
- `copy-webpack-plugin` and `css-minimizer-webpack-plugin` -> `serialize-javascript` 7.0.7 (high, RCE and DoS advisories)
- `sockjs` -> `uuid` 11.1.1 (moderate, GHSA-w5hq-g745-h8pq, via webpack-dev-server); deliberately keyed by sockjs rather than nested under webpack-dev-server, so that any future dependency pulling in sockjs also gets the patched uuid instead of reintroducing the vulnerable range

Scoping keeps unrelated subtrees on their own versions (for example mermaid keeps its own uuid). `markdownlint-cli` itself stays at 0.48.0 because 0.49.x declares `engines.node >=22`.

After this, `npm audit` reports **0 vulnerabilities** and no package in the tree requires Node above 20.

## Testing

- `npm run lint:md` passes with the overridden js-yaml, markdown-it and run-con
- `npm run build:fast` passes
- `docusaurus start` boots and serves HTTP 200 (covers the uuid override in the sockjs dev-server path)

Depends on #556 (branched on top of it, both touch `package-lock.json`).

Part of #628

## Summary by CodeRabbit

* **Chores**
  * Updated dependency resolution settings by adding new override rules to pin specific transitive package versions.
  * Improved consistency and reliability during development and build processes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12755,9 +12755,9 @@
       "license": "MIT"
     },
     "node_modules/js-yaml": {
-      "version": "4.2.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.2.0.tgz",
-      "integrity": "sha512-ePWsvanv0DWuDRsW8dnt+R4jQ31SCRCQ7hhNcPXZPsoBZiemuZNYGf7adZdqX2D86j6rvKp3RpCxVTSb8WQlOw==",
+      "version": "4.3.0",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.3.0.tgz",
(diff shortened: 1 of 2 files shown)


mesutoezdil:
Merged as part of #644 (commit ff549245). The `overrides` block for js-yaml, markdown-it, run-con, serialize-javascript and uuid is on master, so this PR is a no-op now. Closing.

PR #631, opened 2026-07-22
Configure Git-LFS for assets

fishman:
**What type of PR is this?**

/kind feature

The repo is steadily growing and if we start adding videos it will get unmanageable. This PR adds moves assets to git-lfs which is integrated in GitHub.

## Summary by CodeRabbit

* **Chores**
  * Improved handling of large binary assets (images, videos, diagrams, PDFs) by routing them through Git LFS for better storage and consistency.
* **Documentation**
  * Updated the contribution setup guide to include a one-time Git LFS installation and pull step for required image assets.
  * Adjusted the documentation health workflow to enable Git LFS during repository checkout so automated builds can access large files correctly.


mesutoezdil:
thx for pr
pls resolve the conflicts
and check the bots messages

.gitattributes
```diff
@@ -0,0 +1,8 @@
+*.png filter=lfs diff=lfs merge=lfs -text
```
mesutoezdil, line 1: netlify builds dont fetch lfs objects unless GIT_LFS_ENABLED is set in the build env, so rn every img on the deployed site would ship as a tiny pointer file, pls add it to netlify.toml in this pr

.github/workflows/docs-health.yml
```diff
@@ -16,6 +16,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
+        with:
+          lfs: true
```
mesutoezdil, line 21: heads up that gh free lfs gives the org only 1gb of bandwidth per month and every ci run + netlify build will now pull all these imgs, so the quota can run out fast and then lfs downloads fail repo wide

.gitattributes
```diff
@@ -0,0 +1,8 @@
...
+*.jpg filter=lfs diff=lfs merge=lfs -text
+*.jpeg filter=lfs diff=lfs merge=lfs -text
+*.gif filter=lfs diff=lfs merge=lfs -text
+*.drawio filter=lfs diff=lfs merge=lfs -text
+*.mp4 filter=lfs diff=lfs merge=lfs -text
+*.pdf filter=lfs diff=lfs merge=lfs -text
+*.webp filter=lfs diff=lfs merge=lfs -text
```
mesutoezdil, line 8: contributors w/o git-lfs installed will clone pointer files instead of imgs and their new screenshots wont get tracked, imo worth a short note in CONTRIBUTING.md

.github/workflows/docs-health.yml
```diff
@@ -16,6 +16,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
```
mesutoezdil: this branch is behind master which now pins the actions to commit shas (#633), thats why the pr shows a conflict, pls rebase and keep the pinned sha here

.gitattributes
fishman: netlify.toml is read after checkout so that's not a fix, fix was done manually

PR #634, opened 2026-07-22
fix: stop passing GitHub token on the curl command line

mesutoezdil:
## What

In the `fetchFromWebPage()` fallback of `src/scripts/fetch-changelog.js`:

- Drop the `Authorization` header from the fallback fetch. It downloads the public HTML release page, where the API token does nothing except get exposed.
- Replace `execSync` with a string-built shell command by `spawnSync("curl", [args])`. Nothing is interpolated into a shell anymore.
- Use `-fsSL` so HTTP errors fail with a clear message and redirects are followed, add `--max-time 30` so a stalled connection cannot hang the script (the API path already has a 10s timeout), and surface `result.error` and signal terminations explicitly.

## Why

Two problems with the old code:

1. `GH_PAT` was embedded in the shell command line, so it was visible in the process list to any local process and could leak into `execSync` error messages and CI logs.
2. The command string interpolated `GITHUB_URL`, which is built from the CLI `version` argument, into a shell. An argument array removes shell interpretation entirely.

The primary API path already sends the token correctly via HTTPS request headers and is unchanged.

## Testing

- `node --check` and prettier pass
- Success path exercised against the v2.6.0 release page (status 0, full page body returned)
- Error path exercised against a nonexistent tag: curl now exits 56 with "The requested URL returned error: 404" instead of silently returning an empty page


rootsongjc (approved):
/lgtm

PR #635, opened 2026-07-22
chore: use https for adopter and contributor links

mesutoezdil:
## What

Switches two external links from `http://` to `https://`:

- `src/data/adopters.json`: `ragehealth.cn`
- `src/data/contributors.js`: `www.tenxcloud.com`

Both sites were verified to serve HTTP 200 over https. The remaining `http://www.arkunion.com` link is left unchanged because that host does not answer on https.

## Why

LFX Insights control OSPS-BR-03.01 checks that all links in the project use encrypted channels. Plain http links from the site expose visitors to trivial downgrade and injection on the hop.


rootsongjc (approved):
/lgtm

PR #639, opened 2026-07-22
security: add baseline security headers on Netlify

mesutoezdil:
## What

Adds a `for = "/*"` headers block to `netlify.toml` with baseline security headers:

- `X-Content-Type-Options: nosniff` blocks MIME sniffing
- `X-Frame-Options: SAMEORIGIN` blocks clickjacking via third-party framing
- `Referrer-Policy: strict-origin-when-cross-origin` avoids leaking full URLs to external sites
- `Permissions-Policy` disables camera, microphone and geolocation, which the site never uses
- `Strict-Transport-Security` (1 year) keeps browsers on https after the first visit; `includeSubDomains` is deliberately omitted so a future or legacy non-HTTPS subdomain cannot be hard-failed by a policy served from the apex

Existing per-path header blocks are untouched; Netlify merges them with the wildcard block.

## Why

The site currently serves no security headers at all. These five are the standard no-risk baseline for a static documentation site; none of them can break Docusaurus functionality. A CSP is deliberately left out because Docusaurus relies on inline scripts and would need careful nonce handling.

## Testing

- `netlify.toml` parses as valid TOML (10 header blocks)
- The Netlify deploy preview on this PR serves the new headers


rootsongjc (commented):
/lgtm

PR #641, opened 2026-07-22
chore: remove unused npm dependencies

mesutoezdil:
> [!NOTE]
> This PR is stacked on #556 and #630, so until those merge, the "Files changed" diff against master also shows their lockfile refresh and `overrides` block. The only change introduced by this PR itself is the removal of three dependencies; see the last commit for the isolated diff. Merge order: #556 -> #630 -> this, each becoming a trivial diff as its parent lands.

## What

Removes three dependencies that nothing in the repo uses:

- `asciinema-player` (dependency): no import, no CSS reference, no MDX usage anywhere in `src`, `themes`, `docs`, `tutorials`, `blog` or `i18n`
- `react-github-btn` (dependency): the star button is a custom component (`src/components/gitHubButton.js`) built on Font Awesome
- `gh-pages` (devDependency): deployment moved to Netlify (`netlify.toml`); the `docusaurus deploy` script does not use this package

## Why

Every installed package is attack surface for the next supply-chain advisory and noise in `npm audit`. These three contribute nothing to the build.

## Testing

- `npm run build:fast` passes after removal
- `npm audit` still reports 0 vulnerabilities

Part of #628

PR #632, opened 2026-07-22
Add Events page from Google Calendar

fishman:
- Create new Events page with 2-week calendar grid, agenda view, and category filters

Co-authored-by: deepseek <deepseek@users.noreply.github.com>
Signed-off-by: Reza Jelveh <fishmangit@dynamia.ai>

**What type of PR is this?**

/kind documentation
/kind feature

**What this PR does / why we need it**:

This is based on the vLLM events page. https://vllm.ai/events

## Summary by CodeRabbit

* **New Features**
  * Added an Events page accessible from the main navigation.
  * Displays upcoming community events in a two-week calendar and agenda view.
  * Added week navigation, “Today” shortcut, category filters, event details, locations, and tags.
  * Supports recurring events, localized Chinese and English date/time formatting, dark mode, and responsive layouts.
  * Added a community call calendar and a Discord invitation call to action.

PR #643, opened 2026-07-22
chore: resolve all lockfile packages from the official npm registry

mesutoezdil:
> [!NOTE]
> Stacked on #556 -> #630 -> #641; until those merge, their changes appear in this diff too. The change specific to this PR is the registry URL rewrite in `package-lock.json` (last commit).

## What

Rewrites 502 `resolved` URLs in `package-lock.json` from `registry.npmmirror.com` (a third party mirror) to the official `registry.npmjs.org`.

## Why

With mirror URLs in the lockfile, every `npm ci` on CI, Netlify and contributor machines downloads tarballs from a third party host. That adds an unnecessary party to the supply chain and a single foreign point of failure for all builds. The integrity hashes stayed unchanged, so this swaps only the download source, not any package content.

Noticed while reviewing #632, where fresh lockfile entries kept inheriting the mirror URLs.

## Testing

- `rm -rf node_modules && npm ci` succeeds, meaning every package downloaded from the official registry matches the existing sha512 integrity hash
- `npm run build:fast` passes
- Zero `npmmirror` references remain in the lockfile

Part of #628

## Summary by CodeRabbit

* **Chores**
  * Updated the project’s underlying packages and version constraints to support improved maintenance, compatibility, and stability.
  * Removed unused supporting integrations and development tooling.
  * No changes were made to the application’s user-facing features or functionality.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,18 +20,15 @@
         "@fortawesome/free-brands-svg-icons": "^7.3.0",
         "@fortawesome/free-solid-svg-icons": "^7.3.0",
         "@fortawesome/react-fontawesome": "^3.3.1",
-        "asciinema-player": "^3.17.0",
         "clsx": "^1.2.1",
         "node-ical": "^0.26.1",
         "prism-react-renderer": "^2.4.0",
         "react": "^18.2.0",
-        "react-dom": "^18.2.0",
-        "react-github-btn": "^1.4.0"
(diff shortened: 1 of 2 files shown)


mesutoezdil:
Merged as part of #644 (commit ff549245). The lockfile on master resolves every package from registry.npmjs.org with no registry.npmmirror.com entries left, so this PR is a no-op now. Closing.

PR #644, opened 2026-07-22
chore: declare package as private with license and engines

mesutoezdil:
> [!NOTE]
> Stacked on #556 -> #630 -> #641 -> #643; until those merge their changes appear in this diff too. The change specific to this PR is in `package.json` plus the matching lockfile root entry (last commit).

## What

Three metadata fixes in `package.json`:

- `"private": true`: this is a website, not a publishable npm package; the previous `"private": false` with an org-scoped name meant an accidental `npm publish` would upload the whole site source to the public registry
- `"license": "CC-BY-4.0"`: matches the repo `LICENSE` file; the missing field is also why regenerated lockfiles keep churning on the root license entry (seen in #632)
- `"engines": { "node": ">=20.18.1" }`: the minimum Node required by transitive deps in the lockfile (undici, cheerio); CI and Netlify already run recent Node 20, this makes npm warn contributors on older Node instead of failing mysteriously

## Testing

- `npm install` syncs the lockfile root entry cleanly, prettier passes, `npm audit` still 0

Part of #628

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,18 +20,15 @@
         "@fortawesome/free-brands-svg-icons": "^7.3.0",
         "@fortawesome/free-solid-svg-icons": "^7.3.0",
         "@fortawesome/react-fontawesome": "^3.3.1",
-        "asciinema-player": "^3.17.0",
         "clsx": "^1.2.1",
         "node-ical": "^0.26.1",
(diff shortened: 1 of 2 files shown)


rootsongjc (approved):
/lgtm

Issue #561, opened 2026-07-07
Suggestion: Add a lab for running inference of model using a KitOps Modelkit instead of Huggingace


shivaylamba:
**Motivation:** 

The current labs downloads models directly from Hugging Face at runtime. While this is convenient for getting started, many production and enterprise environments require models to be packaged, versioned, and distributed through internal registries rather than fetched from external repositories.

This guide would demonstrate how to replace the runtime Hugging Face download with a KitOps ModelKit. The model is pulled into the Pod via a KitOps initContainer and served locally by vLLM or SGLang, eliminating the need for vllm serve <hf-repo> or runtime Hugging Face downloads.

This approach provides several benefits:

- Demonstrates a production-ready model delivery workflow using OCI artifacts.
- Enables reproducible, versioned model deployments.
- Shows how HAMi integrates with modern AI supply chain tooling beyond the default Hugging Face workflow.

shivaylamba:
This is just a suggestion. Would love to know any feedback from the maintainers, if we can proceed with implementing this. 

This could also be a sub section in either vLLM / SGLang lab as a small - markdown toggle note.

mesutoezdil:
> This is just a suggestion. Would love to know any feedback from the maintainers, if we can proceed with implementing this.
> 
> This could also be a sub section in either vLLM / SGLang lab as a small - markdown toggle note.

good idea! are yu planning to build this 1 yourself, or is this just a suggestion for sb else to pick up? 

if sb take this 1, pls follow: tutorials/_template.md: every command and output should come from a real run, plus a chinese translation, sidebar entry (yeah, we have always an issue), and overview mention

shivaylamba:
Thanks @mesutoezdil yes planning to build one myself.

saiyam1814:
Good one @shivaylamba , looking forward to this lab!

mesutoezdil:
Closing this along with PR #655: the tutorial section is one of the areas #689 is reorganizing, so new labs are on hold until the new structure lands.

If I misread the intent of this change, please say so here and I will reopen it.

mesutoezdil:
reopening along with PR #655.

mesutoezdil:
@shivaylamba any news?

shivaylamba:
Hi @mesutoezdil https://github.com/Project-HAMi/website/pull/655 this is the current PR cc @rudrakshkarpe 

What all other changes would you require?

rudrakshkarpe:
[...]

PR #654, opened 2026-07-23
docs(tutorials): add Lab 15: Run SGLang on HAMi GPU Shares

### LINKED ISSUE #528: Add a lab for Run SGLang on HAMi GPU Shares
https://project-hami.io/tutorials/labs/hami-vllm 

Similar to the vLLM lab, This lab demonstrates how to install HAMi on a Kubernetes cluster that already has NVIDIA GPUs, and use HAMi to schedule SGLang inference service

cc @rudrakshkarpe

rudrakshkarpe:
## What type of PR is this?

/kind documentation

## What this PR does

Adds **Lab 15: Run SGLang on HAMi GPU Shares**, an intermediate tutorial for installing HAMi on an existing NVIDIA GPU cluster and scheduling an SGLang inference service on a HAMi GPU share.

The lab covers:

- checking cluster and GPU prerequisites
- safely handing GPU resource ownership from a vendor NVIDIA device plugin to HAMi
- installing HAMi 2.9.0 with ten explicit GPU shares
- deploying Qwen3-1.7B with pinned SGLang v0.5.7
- testing the OpenAI-compatible API
- verifying the visible 25,000 MiB quota and optional over-quota CUDA behavior
- restoring the original device plugin and preserving pre-existing node labels during cleanup
- equivalent English and Chinese instructions

## Related issue

Fixes #528

## Merge order

This PR should merge before #655. It becomes Lab 15, and the dependent KitOps tutorial becomes Lab 16.

## Validation

- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm test` with 73 tests passing
- [x] English and Chinese production build
- [x] Chinese translation added
- [x] SGLang image pinned and confirmed available
- [x] Commits signed off

Verified environment: kind with an NVIDIA H100 80GB cluster. Original command output was captured on 2026-07-23. The over-quota CUDA allocation added during review is documented as an optional diagnostic and includes a post-test service health check.

## Summary by CodeRabbit

* **Documentation**
  * Added Lab 15, a hands-on tutorial for running SGLang inference services with HAMi GPU sharing.
  * Covers prerequisites, HAMi installation, GPU partitioning, service deployment, API testing, resource verification, troubleshooting, cleanup, and next steps.
  * Added a Chinese-language version and listed the tutorial in the Labs sidebar as intermediate, with an estimated duration of 45 minutes.
  * Improved cleanup instructions to restore existing GPU labels and made quota verification configurable.
  * Added the required DRA RBAC application command for supported clusters.

### DIFF
--- a/i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
+++ b/i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
@@ -138,10 +138,8 @@ kubectl delete daemonset "${VENDOR_PLUGIN_DAEMONSET}" \
 
 ```bash
 GPU_NODE=<gpu-node-name>
+GPU_LABEL_WAS_PRESENT="$(kubectl get node "${GPU_NODE}" -o go-template='{{if hasKey .metadata.labels "gpu"}}true{{else}}false{{end}}')"
 GPU_LABEL_BEFORE="$(kubectl get node "${GPU_NODE}" -o jsonpath='{.metadata.labels.gpu}')"
-if [ "${GPU_LABEL_BEFORE}" != "on" ]; then
-  export HAMI_LAB_ADDED_GPU_LABEL=true
-fi
 kubectl label node "${GPU_NODE}" gpu=on --overwrite
 ```
 
@@ -179,7 +177,15 @@ helm upgrade --install hami hami-charts/hami \
   -n kube-system \
   -f hami-values.yaml \
   --version 2.9.0
+```
 
+> 在部分 ACK Kubernetes 1.36 集群中，如果调度器日志显示 `resource.k8s.io` 权限错误，请应用实验 6 使用的 DRA RBAC 辅助清单：
+>
+> ```bash
+> kubectl apply -f https://raw.githubusercontent.com/Project-HAMi/website/master/tutorials/labs/hami-vllm/hami-scheduler-dra-rbac.yaml
+> ```
+
+```bash
 kubectl rollout status deployment/hami-scheduler -n kube-system
 kubectl rollout status daemonset/hami-device-plugin -n kube-system
 ```
@@ -364,14 +370,19 @@ memory.total [MiB], memory.used [MiB]
 
 容器内 `nvidia-smi` 应显示接近 25,000 MiB 的总显存，而主机仍显示物理卡完整容量。这证明 HAMi 的 NVML 拦截公开了所配置的配额，但还不能单独证明超额 CUDA 分配会失败。
 
-可以选择从单独进程请求一次 26 GiB 分配，以验证 CUDA 层的配额实施。该请求超过 Pod 的完整 25,000 MiB 配额，因此应失败，并且不会保留显存或改变服务进程：
+可以选择从单独进程验证 CUDA 层的配额实施。将 `GPU_QUOTA_MIB` 设置为清单中的 `nvidia.com/gpumem` 值。探针会请求比该配额多 1024 MiB，因此降低其他 GPU 的配额后仍然适用：
 
 ```bash
-kubectl exec -i -n sglang ${POD} -- python3 - <<'PY'
+GPU_QUOTA_MIB=25000
+OVER_QUOTA_MIB=$((GPU_QUOTA_MIB + 1024))
+kubectl exec -i -n sglang ${POD} -- \
+  env OVER_QUOTA_MIB="${OVER_QUOTA_MIB}" python3 - <<'PY'
+import os
 import torch
 
+allocation_mib = int(os.environ["OVER_QUOTA_MIB"])
 try:
-    torch.empty(26 * 1024**3 // 4, dtype=torch.float32, device="cuda")
+    torch.empty(allocation_mib * 1024**2 // 4, dtype=torch.float32, device="cuda")
 except RuntimeError as exc:
     if "out of memory" not in str(exc).lower():
         raise
@@ -381,7 +392,7 @@ else:
 PY
 ```
 
-预期输出包含 `PASS`。随后确认实时服务仍然健康：
+预期输出包含 `PASS`。失败的请求不应保留显存或改变服务进程。随后确认实时服务仍然健康：
 
 ```bash
 curl --fail --silent http://127.0.0.1:8001/health
@@ -422,23 +433,25 @@ kubectl apply -f <original-version-pinned-device-plugin-manifest>
 
 对于 Helm release、GPU Operator 或托管插件，请通过相同的管理机制恢复它，并确认只有一个设备插件 DaemonSet 管理 `nvidia.com/gpu`。
 
-仅当本实验添加了 `gpu=on` 时才移除该标签；请在设置变量的同一 shell 中执行：
+恢复原始 `gpu` 标签状态。请在记录 `GPU_LABEL_WAS_PRESENT` 和 `GPU_LABEL_BEFORE` 的同一 shell 中执行：
 
 ```bash
-if [ "${HAMI_LAB_ADDED_GPU_LABEL:-false}" = true ]; then
+if [ "${GPU_LABEL_WAS_PRESENT}" = true ]; then
+  kubectl label node "${GPU_NODE}" "gpu=${GPU_LABEL_BEFORE}" --overwrite
+else
   kubectl label node "${GPU_NODE}" gpu-
 fi
 ```
 
 ## 验证结果
 
-| 声明                      | 证据                                                   |
-| ------------------------- | ------------------------------------------------------ |
-| HAMi 接管 GPU 调度        | Pod 使用 `hami-scheduler` 并请求 HAMi GPU 资源。       |
-| SGLang 在 HAMi 资源上运行 | Pod Ready，HAMi 注入显存和 SM 限制变量。               |
-| 容器内可见显存配额        | Pod 内显示 25,000 MiB，而主机显示完整物理容量。        |
-| 超额分配被拒绝            | 26 GiB PyTorch CUDA 分配在 25,000 MiB 配额下返回 OOM。 |
-| 推理服务可访问            | 模型列表和聊天接口均正常返回。                         |
+| 声明                      | 证据                                                     |
+| ------------------------- | -------------------------------------------------------- |
+| HAMi 接管 GPU 调度        | Pod 使用 `hami-scheduler` 并请求 HAMi GPU 资源。         |
+| SGLang 在 HAMi 资源上运行 | Pod Ready，HAMi 注入显存和 SM 限制变量。                 |
+| 容器内可见显存配额        | Pod 内显示 25,000 MiB，而主机显示完整物理容量。          |
+| 超额分配被拒绝            | PyTorch CUDA 分配请求比配置配额高 1024 MiB，并返回 OOM。 |
+| 推理服务可访问            | 模型列表和聊天接口均正常返回。                           |
 
 ## 后续步骤
 
--- a/i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
+++ b/i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
@@ -140,8 +140,14 @@ kubectl delete daemonset "${VENDOR_PLUGIN_DAEMONSET}" \
 
 ```bash
 GPU_NODE=<gpu-node-name>
-GPU_LABEL_WAS_PRESENT="$(kubectl get node "${GPU_NODE}" -o go-template='{{if hasKey .metadata.labels "gpu"}}true{{else}}false{{end}}')"
-GPU_LABEL_BEFORE="$(kubectl get node "${GPU_NODE}" -o jsonpath='{.metadata.labels.gpu}')"
+GPU_LABEL_WAS_PRESENT="$(kubectl get node "${GPU_NODE}" -o go-template='{{if .metadata.labels}}{{if index .metadata.labels "gpu"}}true{{else}}false{{end}}{{else}}false{{end}}')" || {
+  echo "无法记录 gpu 标签是否存在" >&2
+  exit 1
+}
+GPU_LABEL_BEFORE="$(kubectl get node "${GPU_NODE}" -o jsonpath='{.metadata.labels.gpu}')" || {
+  echo "无法记录当前的 gpu 标签" >&2
+  exit 1
+}
 kubectl label node "${GPU_NODE}" gpu=on --overwrite
 ```
 
@@ -184,7 +190,13 @@ helm upgrade --install hami hami-charts/hami \
 > 在部分 ACK Kubernetes 1.36 集群中，如果调度器日志显示 `resource.k8s.io` 权限错误，请应用实验 6 使用的 DRA RBAC 辅助清单：
 >
 > ```bash
-> kubectl apply -f https://raw.githubusercontent.com/Project-HAMi/website/master/tutorials/labs/hami-vllm/hami-scheduler-dra-rbac.yaml
+> RBAC_COMMIT="90ac82510bfabe05894dd8037078c59f02a51553"
+> RBAC_SHA256="e0a77f99422230ccc8958aac0d04694347769279ec26a9d4a5ff729f89efe3d9"
+> RBAC_MANIFEST="hami-scheduler-dra-rbac.yaml"
+> curl -fsSLo "${RBAC_MANIFEST}" \
+>   "https://raw.githubusercontent.com/Project-HAMi/website/${RBAC_COMMIT}/tutorials/labs/hami-vllm/hami-scheduler-dra-rbac.yaml"
+> printf '%s  %s\n' "${RBAC_SHA256}" "${RBAC_MANIFEST}" | sha256sum -c -
+> kubectl apply -f "${RBAC_MANIFEST}"
 > ```
 
 ```bash
--- a/i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
+++ b/i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
@@ -32,6 +32,7 @@ toc_max_heading_level: 2
 ## 实验概览
 
 ```mermaid
+%% title: HAMi + SGLang 实验流程图
 flowchart LR
     Step1["步骤 1<br/>检查 GPU 集群"] --> Step2["步骤 2<br/>安装 HAMi"]
     Step2 --> Step3["步骤 3<br/>验证 HAMi 资源"]
@@ -45,6 +46,7 @@ flowchart LR
 ## 部署架构
 
 ```mermaid
+%% title: HAMi + SGLang 部署架构
 flowchart TB
     Client["客户端<br/>curl / OpenAI SDK"] --> PF["端口转发 / LoadBalancer"]
     PF --> SVC["SGLang Service<br/>端口 8001"]
--- a/i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
+++ b/i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
@@ -76,7 +76,22 @@ flowchart TB
 
 ```bash
 kubectl get nodes -o wide
+```
+
+```plaintext
+NAME                      STATUS   ROLES           AGE   VERSION   INTERNAL-IP   OS-IMAGE                       CONTAINER-RUNTIME
+hami-demo-control-plane   Ready    control-plane   2m    v1.36.1   172.19.0.2    Debian GNU/Linux 13 (trixie)   containerd://2.3.1
+```
+
+主机 GPU：
+
+```bash
 nvidia-smi --query-gpu=index,name,memory.total --format=csv
+```
+
+```plaintext
+index, name, memory.total [MiB]
+0, NVIDIA H100 80GB HBM3, 81559 MiB
 ```
 
 安装 HAMi 后，每张物理 GPU 被注册为 10 个可调度共享资源：
@@ -84,6 +99,17 @@ nvidia-smi --query-gpu=index,name,memory.total --format=csv
 ```bash
 kubectl get nodes -o 'custom-columns=NAME:.metadata.name,GPU:.status.allocatable.nvidia\.com/gpu'
 kubectl get pods -n kube-system -l app.kubernetes.io/instance=hami -o wide
+```
+
+```plaintext
+NAME                      GPU
+hami-demo-control-plane   10
+```
+
+```plaintext
+NAME                              READY   STATUS    NODE
+hami-device-plugin-...            2/2     Running   hami-demo-control-plane
+hami-scheduler-...                2/2     Running   hami-demo-control-plane
 ```
 
 ## 步骤 1：检查 GPU 集群
@@ -95,15 +121,13 @@ kubectl describe node | grep -A8 -E "Capacity:|Allocatable:" | grep -E "nvidia.c
 
 HAMi 和厂商 NVIDIA 设备插件会注册相同的 `nvidia.com/gpu` 资源，因此二者不能同时运行在同一 GPU 节点。保留 NVIDIA 驱动、Container Toolkit 和运行时，但在安装 HAMi 前禁用或移除厂商设备插件。
 
-在自管理集群中，先找到、备份并移除现有 DaemonSet：
+先确认现有插件的安装方式。稍后需要使用原始 Helm release、Operator 配置、托管插件设置或源清单来恢复它：
 
 ```bash
 kubectl get daemonsets --all-namespaces | grep -E 'nvidia.*device-plugin'
 
 VENDOR_PLUGIN_NAMESPACE=<namespace>
 VENDOR_PLUGIN_DAEMONSET=<daemonset-name>
-kubectl get daemonset "${VENDOR_PLUGIN_DAEMONSET}" \
-  -n "${VENDOR_PLUGIN_NAMESPACE}" -o yaml >nvidia-device-plugin-backup.yaml
 kubectl delete daemonset "${VENDOR_PLUGIN_DAEMONSET}" \
   -n "${VENDOR_PLUGIN_NAMESPACE}"
 ```
@@ -137,22 +161,18 @@ global:
     gpu: "on"
 
 devicePlugin:
+  # 这是 chart 的默认值。此处显式展示，因为实验会验证 10 个共享资源。
   deviceSplitCount: 10
 
 scheduler:
-  # 与集群 Kubernetes 版本一致；示例为 kind v1.36.1。
-  kubeScheduler:
-    image:
-      tag: "v1.36.1"
   leaderElect: false
 ```
 
-| 配置                                   | 说明                                             |
-| -------------------------------------- | ------------------------------------------------ |
+| 配置 | 说明 |
+| --- | --- |
 | `global.managedNodeSelector.gpu: "on"` | 仅在带 `gpu=on` 标签的节点上运行 HAMi 设备插件。 |
-| `devicePlugin.deviceSplitCount: 10`    | 将每张物理 GPU 注册为 10 个 vGPU。               |
-| `scheduler.kubeScheduler.image.tag`    | 必须与集群 Kubernetes 版本一致。                 |
-| `scheduler.leaderElect: false`         | 本实验使用单副本调度器。                         |
+| `devicePlugin.deviceSplitCount: 10` | 将每张物理 GPU 注册为 10 个 vGPU。该值与当前 chart 默认值一致，并因实验会验证结果而显式设置。 |
+| `scheduler.leaderElect: false` | 本实验使用单副本调度器。 |
 
 ```bash
 helm upgrade --install hami hami-charts/hami \
@@ -291,7 +311,7 @@ kubectl -n sglang port-forward svc/sglang-qwen3-17b 8001:8001
 ## 步骤 6：测试推理
 
 ```bash
-curl -s http://127.0.0.1:8001/v1/models | jq
+curl -s http://127.0.0.1:8001/v1/models | python3 -m json.tool
 
 curl -s http://127.0.0.1:8001/v1/chat/completions \
   -H 'Content-Type: application/json' \
@@ -301,7 +321,7 @@ curl -s http://127.0.0.1:8001/v1/chat/completions \
     "temperature": 0.2,
     "max_tokens": 80,
     "chat_template_kwargs": {"enable_thinking": false}
-  }' | jq
+  }' | python3 -m json.tool
 ```
 
 若响应包含 `choices[0].message.content`，说明 SGLang 推理服务正常工作。
@@ -316,9 +336,35 @@ kubectl exec -n sglang ${POD} -- env | grep -E 'CUDA_DEVICE|NVIDIA_VISIBLE'
 kubectl exec -n sglang ${POD} -- nvidia-smi
 ```
 
+HAMi 注入的环境变量示例：
+
+```plaintext
+NVIDIA_VISIBLE_DEVICES=GPU-04b76a6c-da10-342f-e9f5-5f5684eacb86
+CUDA_DEVICE_MEMORY_LIMIT_0=25000m
+CUDA_DEVICE_SM_LIMIT=30
+```
+
+Pod 内的预期显存视图：
+
+```plaintext
+| GPU  Name                 ... | Memory-Usage          |
+| NVIDIA H100 80GB HBM3     ... | 18213MiB / 25000MiB   |
+```
+
+主机仍显示完整物理容量：
+
+```bash
+nvidia-smi --query-gpu=memory.total,memory.used --format=csv
+```
+
+```plaintext
+memory.total [MiB], memory.used [MiB]
+81559 MiB, 18560 MiB
+```
+
 容器内 `nvidia-smi` 应显示接近 25,000 MiB 的总显存，而主机仍显示物理卡完整容量。这证明 HAMi 的 NVML 拦截公开了所配置的配额，但还不能单独证明超额 CUDA 分配会失败。
 
-请求一次 26 GiB 分配，以验证 CUDA 层的配额实施：
+可以选择从单独进程请求一次 26 GiB 分配，以验证 CUDA 层的配额实施。该请求超过 Pod 的完整 25,000 MiB 配额，因此应失败，并且不会保留显存或改变服务进程：
 
 ```bash
 kubectl exec -i -n sglang ${POD} -- python3 - <<'PY'
@@ -335,7 +381,14 @@ else:
 PY
 ```
 
-预期输出包含 `PASS`。HAMi 通过拦截 CUDA 分配调用实施软件显存配额，这不同于 MIG 等硬件分区；算力限制同样属于软件节流。
+预期输出包含 `PASS`。随后确认实时服务仍然健康：
+
+```bash
+curl --fail --silent http://127.0.0.1:8001/health
+curl --fail --silent http://127.0.0.1:8001/v1/models | python3 -m json.tool
+```
+
+HAMi 通过拦截 CUDA 分配调用实施软件显存配额，这不同于 MIG 等硬件分区；算力限制同样属于软件节流。
 
 ## 故障排除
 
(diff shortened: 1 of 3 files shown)


tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,547 @@
...
+    HAMI_D --> N1
+    POD --> N1
+```
+
+## Prerequisites
+
+You need to prepare in advance:
+
+- A working Kubernetes cluster
+- At least 1 NVIDIA GPU node with enough free memory for the model (this guide uses a single NVIDIA H100 80GB; an A10/L40S-class GPU also works if you keep the small model and `gpumem` values)
```
mesutoezdil, line 73: an a10 is only ~23028 MiB, so with `gpumem: 25000` below the pod just sits pending, so either drop the a10 from this line or give an a10 specific gpumem value?

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,547 @@
...
+helm repo update hami-charts
+```
+
+Create a values file (save as `hami-values.yaml`):
+
+```yaml
+device:
+  nvidia:
+    driver:
+      enabled: false
```
mesutoezdil, line 169: `device.nvidia.driver.enabled` is not a real key in hami 2.9.0, there is no top level `device:` in the chart and no template reads `.Values.device`, so helm drops the whole block without a warning (no values.schema.json either)?

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,547 @@
...
+  managedNodeSelector:
+    gpu: "on"
+
+devicePlugin:
+  deviceSplitCount: 10
+
+scheduler:
+  # Match your cluster Kubernetes minor version (example: kind v1.36.1).
+  kubeScheduler:
+    imageTag: "v1.36.1"
```
mesutoezdil, line 182: the chart reads `scheduler.kubeScheduler.image.tag`, not `imageTag`, so this is ignored and the tag falls back to the auto detected k8s version, which makes the table row below misleading?

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,547 @@
...
+      labels:
+        app.kubernetes.io/name: sglang-qwen3-17b
+      annotations:
+        hami.io/node-scheduler-policy: binpack
+        hami.io/gpu-scheduler-policy: binpack
+    spec:
+      schedulerName: hami-scheduler
+      containers:
+        - name: sglang
+          image: lmsysorg/sglang:latest
```
mesutoezdil, line 274: pls pin a tag or digest, this would be the only `:latest` under `tutorials/` and combined with `imagePullPolicy: IfNotPresent` two nodes can end up on diff builds, which does not match the `verified: 2026-07-23` claim

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,547 @@
...
+          command:
+            - python3
+            - -m
+            - sglang.launch_server
+            - --model-path=Qwen/Qwen3-1.7B
+            - --host=0.0.0.0
+            - --port=30000
+            - --mem-fraction-static=0.7
+            - --context-length=8192
+            - --attention-backend=triton
```
mesutoezdil, line 285: worth one row in the key points table on why triton and not the default backend, otherwise ppl on other gpus copy it blind

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,547 @@
...
+  ports:
+    - name: http
+      port: 8001
+      targetPort: http
+EOF
+```
+
+Key points:
+
+| Configuration | Description |
```
mesutoezdil, line 337: `prettier --check` fails on this table only, so run `npm run format` before push or docs-health ci will block the pr

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,547 @@
...
+| GPU node runs HAMi device plugin | `hami-device-plugin` is Ready and advertises `nvidia.com/gpu=10`. |
+| SGLang runs on HAMi resources | Pod Ready; HAMi injects `CUDA_DEVICE_MEMORY_LIMIT_0=25000m` and `CUDA_DEVICE_SM_LIMIT=30`. |
+| Memory cap is visible in-container | In-pod `nvidia-smi` shows `... / 25000MiB` while host still shows `81559 MiB`. |
+| Inference service is accessible | `/v1/models` returns `Qwen/Qwen3-1.7B`; chat endpoint returns content. |
+
+## Next Steps
+
+- Increase `replicas` and observe how HAMi packs multiple SGLang Pods with `binpack`.
+- Lower `nvidia.com/gpumem` / `nvidia.com/gpucores` further and co-locate another small workload on the same GPU.
+- Deliver the model from an OCI registry instead of Hugging Face: a companion **KitOps ModelKit** lab replaces the runtime download with a `kitops-init` initContainer ([website#561](https://github.com/Project-HAMi/website/issues/561)).
```
mesutoezdil, line 546: #655 already adds this lab, so point at the lab page instead of issue #561 once it merges.

sidebars-tutorials.js
```diff
@@ -61,6 +61,11 @@ module.exports = {
           id: "labs/topology-aware-scheduling",
           customProps: { level: "Intermediate", duration: "about 45 minutes" },
         },
+        {
+          type: "doc",
```
mesutoezdil, line 65: #655 appends to this same list off master, so whichever merges second needs a rebase and the lab number depends on that order.

mesutoezdil (commented):
nice lab, reads well next to lab 6. left a few notes inline. the one i'd not skip is the zh mirror: the other 10 labs all have a file under `i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/`, incl lab 10 which landed recently, so without it the zh site quietly serves this one page in english. build passes locally (`npm run build:fast`) and links are fine.

mesutoezdil:
and pls check the revieews of coderrabbit

tutorials/labs/hami-sglang.md
Creativeklvn, line 274: The SGLang image is currently pinned to `lmsysorg/sglang:latest`. Would it be better to pin a specific SGLang version for reproducibility? Using `latest` could cause the lab to behave differently or break if a future SGLang release changes the image or CLI behavior.

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,547 @@
+---
+title: "Lab 11: Run SGLang on HAMi GPU Shares"
```
mesutoezdil, line 2: lab 11 is taken on master. renumber to lab 14 here, in sidebar_label and pr title.

sidebars-tutorials.js
```diff
@@ -61,6 +61,11 @@ module.exports = {
...
+          id: "labs/hami-sglang",
```
mesutoezdil, line 66: move last, after labs/volcano-ascend-vnpu.

tutorials/labs/hami-sglang.md
mesutoezdil, line 274: pin a version tag, not latest.

mesutoezdil (commented):
checklist is empty. rebase, run checks, mark ready.

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,547 @@
...
+This guide is modeled after [Lab 6: Run vLLM on HAMi GPU Shares](./hami-vllm). The steps are not tied to a specific cloud vendor. As long as your Kubernetes cluster has available NVIDIA GPUs, NVIDIA drivers, and container runtime support, you can reproduce the same setup.
+
+## Learning Objectives
+
+- Verify that an existing GPU Kubernetes cluster meets the prerequisites for HAMi and SGLang
+- Install HAMi scheduler and device plugin
+- Add labels required by the HAMi DaemonSet to GPU nodes
+- Run SGLang using HAMi's `nvidia.com/gpu`, `nvidia.com/gpumem`, and `nvidia.com/gpucores` resources
+- Test the SGLang OpenAI-compatible API via port forwarding
+- Confirm that `nvidia.com/gpumem` is enforced inside the SGLang Pod
```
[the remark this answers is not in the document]
rudrakshkarpe: Fixed in f58d7b03. The lab now distinguishes HAMi's software CUDA/NVML enforcement from hardware partitioning, retains the in-Pod quota visibility check as supporting evidence, and adds a 26 GiB PyTorch CUDA allocation that must return an out-of-memory error under the 25,000 MiB quota. The same validation is included in the Chinese mirror.

tutorials/labs/hami-sglang.md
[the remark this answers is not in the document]
rudrakshkarpe: Fixed in f58d7b03. The default path now requires more than 25,000 MiB of usable VRAM. A10 users are explicitly told to lower every `gpumem` request and limit below the memory reported by `nvidia-smi`; the architecture diagram no longer implies the default manifest fits an A10.

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,547 @@
...
+        app.kubernetes.io/name: sglang-qwen3-17b
+      annotations:
+        hami.io/node-scheduler-policy: binpack
+        hami.io/gpu-scheduler-policy: binpack
+    spec:
+      schedulerName: hami-scheduler
+      containers:
+        - name: sglang
+          image: lmsysorg/sglang:latest
+          imagePullPolicy: IfNotPresent
```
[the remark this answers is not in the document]
rudrakshkarpe: Fixed in f58d7b03 in both language manifests. The image is pinned to the existing `lmsysorg/sglang:v0.5.7` release tag and retains `IfNotPresent`; I also verified that the tag is published.

tutorials/labs/hami-sglang.md
rudrakshkarpe: Fixed in f58d7b03. The default manifest now requires more than 25,000 MiB usable VRAM, with explicit A10 instructions to lower both requests and limits below the card's reported capacity.

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,547 @@
...
+kubectl describe node | grep -A8 -E "Capacity:|Allocatable:" | grep -E "nvidia.com/gpu|cpu:|memory:"
+```
+
+If the cluster already has a vendor NVIDIA device plugin installed, you may already see `nvidia.com/gpu`. After installing HAMi, `nvidia.com/gpu` becomes the number of vGPUs exposed by HAMi.
+
+HAMi's device plugin matches `gpu=on` when managed node selectors are enabled. Label your GPU nodes:
+
+```bash
+kubectl label node <gpu-node-name> gpu=on --overwrite
+```
```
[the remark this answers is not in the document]
rudrakshkarpe: Fixed in f58d7b03 in both languages. The guide now requires the vendor NVIDIA device plugin to be disabled/removed before HAMi while retaining drivers/runtime, distinguishes self-managed from operator-managed clusters, saves a self-managed DaemonSet manifest, and restores the vendor plugin after HAMi cleanup.

tutorials/labs/hami-sglang.md
rudrakshkarpe: Fixed in f58d7b03. The nonexistent `device.nvidia.driver.enabled` block and its misleading table row were removed. Driver/runtime prerequisites are now documented separately from the chart values.

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,547 @@
...
+
+```bash
+helm uninstall hami -n kube-system
+```
+
+Optional label cleanup:
+
+```bash
+kubectl label node <gpu-node-name> gpu-
+```
```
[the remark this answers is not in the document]
rudrakshkarpe: Fixed in f58d7b03 in both languages. The setup records whether `gpu=on` already existed, sets `HAMI_LAB_ADDED_GPU_LABEL` only when this lab changed the node, and cleanup conditionally removes the label from the same shell.

tutorials/labs/hami-sglang.md
rudrakshkarpe: Fixed in f58d7b03. The deployment is pinned to `lmsysorg/sglang:v0.5.7` instead of `latest`, and the Chinese mirror uses the same tag.

tutorials/labs/hami-sglang.md
rudrakshkarpe: Fixed. Prettier was run on the English lab, Chinese mirror, and sidebar; the targeted `prettier --check` now passes.

tutorials/labs/hami-sglang.md
rudrakshkarpe: Fixed in f58d7b03. The values file and explanatory table now use the actual HAMi 2.9.0 path: `scheduler.kubeScheduler.image.tag`.

tutorials/labs/hami-sglang.md
rudrakshkarpe: Added in f58d7b03. The key-points table explains that Triton is the backend verified on the H100 environment and tells readers on other GPU architectures to select a backend supported by their SGLang version and hardware.

tutorials/labs/hami-sglang.md
rudrakshkarpe: Rebased onto current `master`; because master now contains Labs 1–14, this is renumbered to Lab 15 in the English and Chinese frontmatter and the PR title.

tutorials/labs/hami-sglang.md
rudrakshkarpe: Fixed in f58d7b03. The image is now pinned to `lmsysorg/sglang:v0.5.7` in both language versions.

tutorials/labs/hami-sglang.md
rudrakshkarpe: Updated in f58d7b03. The next-step link now points directly to companion PR #655 rather than issue #561. A relative lab-page link would fail until #655 merges, so the PR link remains valid during either merge order.

sidebars-tutorials.js
rudrakshkarpe: Resolved during the rebase. The SGLang entry is now last in the current sidebar, after the newer Lab 14 (`composable-scheduler-policies-gke`).

tutorials/labs/hami-sglang.md
rudrakshkarpe: Agreed and fixed in f58d7b03: `lmsysorg/sglang:v0.5.7` replaces `latest` in both English and Chinese manifests.

sidebars-tutorials.js
rudrakshkarpe: Resolved by rebasing onto current `master`. SGLang is now the last sidebar entry and Lab 15. The companion KitOps PR will follow as Lab 16, preserving the intended learning order.

rudrakshkarpe:
All review feedback is addressed in f58d7b03 after rebasing onto current `master`:

- renumbered the lab to Lab 15 and moved it to the end of the current sidebar
- added the missing Chinese mirror with matching safeguards and deployment settings
- pinned SGLang to `lmsysorg/sglang:v0.5.7`
- corrected the HAMi 2.9.0 Helm value to `scheduler.kubeScheduler.image.tag` and removed the nonexistent driver key
- documented safe migration away from—and restoration of—the vendor NVIDIA device plugin
- corrected A10/25,000 MiB guidance
- preserved pre-existing `gpu=on` labels during cleanup
- clarified software CUDA/NVML enforcement and added an over-quota CUDA OOM test
- documented why Triton was used on the verified H100 environment
- linked directly to companion KitOps PR #655

Validation completed:
- `npm run lint`
- `npm test` (73/73)
- Prettier check for the English lab, Chinese mirror, and sidebar
- confirmed the pinned SGLang v0.5.7 image exists
- production build completed for both `en` and `zh`
- signed-off commit and DCO pass

All 17 existing review threads have targeted replies and are resolved. @mesutoezdil, could you please re-review for LGTM when CI is green?

mesutoezdil:
Closing this for now: the tutorial section is one of the areas #689 is reorganizing, so new labs are on hold until the new structure lands.

shivaylamba:
so should we reopen this once the tutorial section is re-opened? @mesutoezdil

mesutoezdil:
> so should we reopen this once the tutorial section is re-opened? @mesutoezdil

sure!

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,593 @@
...
+If this cluster is only used for this lab, you can also uninstall HAMi:
+
+```bash
+helm uninstall hami -n kube-system
+```
+
+Restore the vendor device plugin after HAMi is removed. For the self-managed example above:
+
+```bash
+kubectl apply -f nvidia-device-plugin-backup.yaml
```
mesutoezdil: line 147 saves the daemonset with `kubectl get -o yaml`, so the file carries resourceVersion, uid and status. apply rejects that on a create. did the restore step run on the verification cluster?

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,593 @@
...
+    gpu: "on"
+
+devicePlugin:
+  deviceSplitCount: 10
+
+scheduler:
+  # Match your cluster Kubernetes minor version (example: kind v1.36.1).
+  kubeScheduler:
+    image:
+      tag: "v1.36.1"
```
mesutoezdil: the 2.9.0 chart falls back to the cluster version when this tag is empty, and #712 removes the same flag from the install guide. is pinning it still needed here?

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,593 @@
...
+Create a values file (save as `hami-values.yaml`):
+
+```yaml
+global:
+  managedNodeSelectorEnable: true
+  managedNodeSelector:
+    gpu: "on"
+
+devicePlugin:
+  deviceSplitCount: 10
```
mesutoezdil, line 186: 10 is already the chart default for deviceSplitCount. is it in the values file to make the 10 shares explicit for the reader, or is something else expected to change it?

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,593 @@
...
+That contrast confirms that HAMi's NVML interception exposes the configured quota. It is supporting evidence, but it does not by itself prove that an over-quota CUDA allocation is rejected.
+
+Verify CUDA-level enforcement by requesting a single 26 GiB allocation, which exceeds this Pod's 25,000 MiB quota:
+
+```bash
+kubectl exec -i -n sglang ${POD} -- python3 - <<'PY'
+import torch
+
+try:
+    torch.empty(26 * 1024**3 // 4, dtype=torch.float32, device="cuda")
```
mesutoezdil: sglang has already taken about 18 GiB of the 25000 MiB quota by this point, so only about 7 GiB is free. does this run against the live server without disturbing it?

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,593 @@
+---
+title: "Lab 15: Run SGLang on HAMi GPU Shares"
```
mesutoezdil, line 2: correction on my earlier note: master already has Lab 14, so Lab 15 here is right. the pr body still says Lab 11, no chinese yet and draft, while the diff has all three. could the body be refreshed?

i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
```diff
@@ -0,0 +1,395 @@
...
+
+## 步骤 7：检查 GPU 分配
+
+```bash
+POD=$(kubectl get pod -n sglang -l app.kubernetes.io/name=sglang-qwen3-17b -o jsonpath='{.items[0].metadata.name}')
+kubectl get pod -n sglang ${POD} \
+  -o jsonpath='{.spec.schedulerName}{"\n"}{.spec.containers[0].resources.limits}{"\n"}'
+kubectl exec -n sglang ${POD} -- env | grep -E 'CUDA_DEVICE|NVIDIA_VISIBLE'
+kubectl exec -n sglang ${POD} -- nvidia-smi
+```
```
mesutoezdil, line 357: the english step 7 prints the injected env vars, the in pod 25000MiB line and the host 81559MiB line. none of them are here, so a chinese reader has nothing to compare against. was that on purpose?

i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
```diff
@@ -0,0 +1,395 @@
...
+```bash
+kubectl -n sglang port-forward svc/sglang-qwen3-17b 8001:8001
+```
+
+在另一个终端中执行后续测试。
+
+## 步骤 6：测试推理
+
+```bash
+curl -s http://127.0.0.1:8001/v1/models | jq
```
mesutoezdil: english uses python3 -m json.tool here and jq is not in the prerequisites. should this use the same command?

i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
```diff
@@ -0,0 +1,395 @@
...
+> **kind 提示：** 使用 GPU 直通时，需在 kind 节点内配置 `nvidia-container-toolkit`，并在安装 HAMi 前将 containerd 的 `default_runtime_name` 设为 `nvidia`。
+
+## 示例集群状态
+
+本指南在由一张 NVIDIA H100 80GB 支持的 kind 控制平面节点上完成验证。
+
+```bash
+kubectl get nodes -o wide
+nvidia-smi --query-gpu=index,name,memory.total --format=csv
+```
```
mesutoezdil, line 92: the english page shows the node table, the 81559 MiB card and the allocatable 10 under this section. the chinese one only has the commands. same question as step 7.

tutorials/labs/hami-sglang.md
mesutoezdil: one more on this line: the aliyun mirror often lags upstream kubernetes. does registry.cn-hangzhou.aliyuncs.com carry a kube-scheduler:v1.36.1 tag, or would a reader on 1.36 hit ImagePullBackOff?

i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
rudrakshkarpe: Agreed. Fixed in d0466a77 by replacing both Chinese `jq` pipelines with `python3 -m json.tool`, matching the English page without adding another prerequisite.

i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
rudrakshkarpe: Not intentional. Fixed in d0466a77. The Chinese Step 7 now includes the injected HAMi variables, the in-Pod 25,000 MiB view, the host 81,559 MiB view, the optional OOM result, and the post-test service health check.

tutorials/labs/hami-sglang.md
rudrakshkarpe: Yes, it is intentionally explicit for the reader. Although 10 is the current chart default, the lab later verifies `nvidia.com/gpu=10`. I added a comment and table explanation in d0466a77 that state this is the default and is shown so the verified behavior is clear and easy to customize.

i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
rudrakshkarpe: Not intentional. Fixed in d0466a77. The Chinese example cluster section now includes the node table, H100 81,559 MiB capacity, ten allocatable HAMi shares, and HAMi component status.

tutorials/labs/hami-sglang.md
rudrakshkarpe: Agreed. Fixed in d0466a77 by removing the explicit kube-scheduler image tag and its table row. HAMi 2.9.0 can select the cluster-compatible version, which also avoids depending on a matching tag being available in the configured registry mirror.

tutorials/labs/hami-sglang.md
rudrakshkarpe: Fixed. I replaced the stale PR body with the current Lab 15 scope, Chinese mirror status, validation results, ready-for-review state, and the required merge order before #655.

tutorials/labs/hami-sglang.md
rudrakshkarpe: Good catch. The restore path was not exercised from that live-object export, and the export is not a portable installation manifest because it contains server-managed fields. Fixed in d0466a77: the lab now tells readers to restore the plugin through its original version-pinned manifest, Helm release, Operator, or managed add-on instead of applying a `kubectl get -o yaml` backup.

tutorials/labs/hami-sglang.md
rudrakshkarpe: The allocation runs in a separate process and requests more than the entire 25,000 MiB quota, so a failed request should not retain memory or alter the serving process. This exact diagnostic was added after the original verification run, so d0466a77 now labels it optional, explains the isolation, and immediately checks `/health` and `/v1/models` afterward to confirm the live server remains healthy.

tutorials/labs/hami-sglang.md
rudrakshkarpe: That registry lag is a real portability risk. Fixed in d0466a77 by removing the explicit v1.36.1 scheduler image tag in both languages and relying on the HAMi 2.9.0 chart's cluster-version selection.

rudrakshkarpe:
Maintainer review update in d0466a77:

- removed the non-portable live DaemonSet export and restore flow
- documented restoration through the original manifest, Helm release, Operator, or managed add-on
- removed the explicit kube-scheduler image tag and registry-mirror dependency
- retained `deviceSplitCount: 10` with an explanation that it matches the current default
- made the CUDA OOM diagnostic optional and added post-test health checks
- completed the missing Chinese expected outputs
- replaced undeclared `jq` usage with `python3 -m json.tool`
- refreshed the complete PR description

Validation remains green locally for Markdown lint, Prettier, and 73 tests. The signed-off commit is pushed and DCO passes. All nine new maintainer threads have replies and are resolved.

@mesutoezdil, please review for LGTM when CI completes. This PR should merge before #655.

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,594 @@
...
+That contrast confirms that HAMi's NVML interception exposes the configured quota. It is supporting evidence, but it does not by itself prove that an over-quota CUDA allocation is rejected.
+
+Optionally verify CUDA-level enforcement by requesting a single 26 GiB allocation from a separate process. The request exceeds this Pod's entire 25,000 MiB quota, so it should fail without retaining memory or changing the serving process:
+
+```bash
+kubectl exec -i -n sglang ${POD} -- python3 - <<'PY'
+import torch
+
+try:
+    torch.empty(26 * 1024**3 // 4, dtype=torch.float32, device="cuda")
```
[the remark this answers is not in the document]
rudrakshkarpe: Fixed in a7f5af6d in both languages. The probe now derives its request from `GPU_QUOTA_MIB` and allocates 1024 MiB above the configured quota. Readers who lower `gpumem` for an A10 can set the matching quota value instead of using a fixed 26 GiB allocation.

i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
```diff
@@ -0,0 +1,448 @@
...
+
+```bash
+helm upgrade --install hami hami-charts/hami \
+  -n kube-system \
+  -f hami-values.yaml \
+  --version 2.9.0
+
+kubectl rollout status deployment/hami-scheduler -n kube-system
+kubectl rollout status daemonset/hami-device-plugin -n kube-system
+```
```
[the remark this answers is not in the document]
rudrakshkarpe: Fixed in a7f5af6d. The Chinese installation section now includes the same ACK Kubernetes 1.36 DRA RBAC recovery command as the English page, placed after Helm installation and before rollout verification.

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,594 @@
...
+
+HAMi's device plugin matches `gpu=on` when managed node selectors are enabled. Record whether the label already existed, then label the GPU node:
+
+```bash
+GPU_NODE=<gpu-node-name>
+GPU_LABEL_BEFORE="$(kubectl get node "${GPU_NODE}" -o jsonpath='{.metadata.labels.gpu}')"
+if [ "${GPU_LABEL_BEFORE}" != "on" ]; then
+  export HAMI_LAB_ADDED_GPU_LABEL=true
+fi
+kubectl label node "${GPU_NODE}" gpu=on --overwrite
```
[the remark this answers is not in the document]
rudrakshkarpe: Fixed in a7f5af6d in both languages. The single-node flow now records whether the `gpu` label existed and its exact prior value, then cleanup restores that value or removes the label only when it was originally absent. I removed the optional ACK bulk-label shortcut so the lab does not modify multiple nodes without per-node restoration state.

rudrakshkarpe:
Follow-up review update in a7f5af6d: the Chinese ACK DRA recovery command is now present, original non-`on` GPU label values are restored exactly, the untracked ACK bulk-label shortcut was removed, and the CUDA OOM probe is derived from the configured quota. All three follow-up CodeRabbit threads have replies and are resolved.

mesutoezdil:
/lgtm

i checked all nine of my findings against a7f5af6d rather than the summary. the two that mattered are properly fixed:

- the restore step no longer hands `kubectl apply` a live `get -o yaml` export, and the text now names that trap directly
- the over quota test derives the size from the actual quota instead of a fixed 26 GiB, and it now confirms the server is still healthy afterwards

the deviceSplitCount answer is a fair one, and the chinese page carries the captured evidence now, including the 25000MiB against 81559MiB contrast and the PASS line.

one small thing left for a follow up, not blocking: the chinese page still has seven output blocks against eleven in english. missing are the step 2 rollout, the step 4 pod listing and scheduler events, and the step 6 models response. worth closing that gap when you touch the file again.

i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
```diff
@@ -0,0 +1,461 @@
...
+
+- 检查 GPU Kubernetes 集群是否满足 HAMi 和 SGLang 的前置条件
+- 安装 HAMi 调度器和设备插件
+- 使用 `nvidia.com/gpu`、`nvidia.com/gpumem` 和 `nvidia.com/gpucores` 运行 SGLang
+- 通过端口转发测试兼容 OpenAI 的 API
+- 验证容器内可见配额以及超额 CUDA 分配被拒绝
+
+## 实验概览
+
+```mermaid
```
rootsongjc, line 34: These two mermaids are missing titles.

rootsongjc (changes requested):

i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/hami-sglang.md
rudrakshkarpe: Good catch. I added titles to both Mermaid diagrams in the Chinese page and verified the full site build. Thanks!

rudrakshkarpe:
Hi @rootsongjc, I added the missing titles to both Mermaid diagrams and verified the full site build. Could you please take another look? Thanks!

mesutoezdil:
/lgtm cancel

my mistake, i gave lgtm too early. the bot is right about line 156, and it is in the very line i praised.

`hasKey` is a sprig function, not a Go `text/template` builtin, and `kubectl -o go-template` uses plain `text/template`. i compiled that exact template and Go rejects it at parse time:

```
template: x:1: function "hasKey" not defined
```

so the cleanup step fails outright rather than restoring the label. `index` is the builtin that does this job:

```bash
GPU_LABEL_WAS_PRESENT="$(kubectl get node "${GPU_NODE}" -o go-template='{{if .metadata.labels}}{{if index .metadata.labels "gpu"}}true{{else}}false{{end}}{{else}}false{{end}}')"
```

the line 206 point also looks fair, that manifest is pulled from the mutable `master` branch, so what a reader applies today is not what you validated. a commit sha would pin it.

on line 500 i am happy with what you have. deriving the size from the quota and checking `/health` afterwards answers the concern i raised, and a separate pod would need its own gpu share.

happy to lgtm again once 156 is fixed.

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,594 @@
...
+  -n "${VENDOR_PLUGIN_NAMESPACE}"
+```
+
+> If a GPU Operator or managed Kubernetes add-on owns the DaemonSet, disable that component through its operator/add-on configuration instead; otherwise its controller may recreate the DaemonSet. Do not uninstall the host driver or NVIDIA container runtime.
+
+HAMi's device plugin matches `gpu=on` when managed node selectors are enabled. Record whether the label already existed, then label the GPU node:
+
+```bash
+GPU_NODE=<gpu-node-name>
+GPU_LABEL_WAS_PRESENT="$(kubectl get node "${GPU_NODE}" -o go-template='{{if hasKey .metadata.labels "gpu"}}true{{else}}false{{end}}')"
```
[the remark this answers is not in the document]
rudrakshkarpe: Fixed. I replaced hasKey with the supported index function and added error checks for both label snapshots. I made the same change in the Chinese page.

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,594 @@
...
+helm upgrade --install hami hami-charts/hami \
+  -n kube-system \
+  -f hami-values.yaml \
+  --version 2.9.0
+```
+
+> On some ACK Kubernetes 1.36 clusters, the built-in kube-scheduler in HAMi also needs DRA-related RBAC. If the scheduler logs show `resource.k8s.io` permission errors, apply the helper used in [Lab 6](./hami-vllm):
+>
+> ```bash
+> kubectl apply -f https://raw.githubusercontent.com/Project-HAMi/website/master/tutorials/labs/hami-vllm/hami-scheduler-dra-rbac.yaml
```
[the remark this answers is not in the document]
rudrakshkarpe: Fixed. The RBAC manifest now uses an immutable commit and verifies its SHA-256 checksum before applying it. I made the same change in the Chinese page.

tutorials/labs/hami-sglang.md
```diff
@@ -0,0 +1,594 @@
...
+```
+
+That contrast confirms that HAMi's NVML interception exposes the configured quota. It is supporting evidence, but it does not by itself prove that an over-quota CUDA allocation is rejected.
+
+Optionally verify CUDA-level enforcement from a separate process. Set `GPU_QUOTA_MIB` to the `nvidia.com/gpumem` value used in your manifest. The probe requests 1024 MiB more than that quota, so it remains valid if you lower the quota for another GPU:
+
+```bash
+GPU_QUOTA_MIB=25000
+OVER_QUOTA_MIB=$((GPU_QUOTA_MIB + 1024))
+kubectl exec -i -n sglang ${POD} -- \
```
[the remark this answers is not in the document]
rudrakshkarpe: Thanks. I kept the current probe based on the maintainer confirmation that deriving the size from the quota and checking service health addresses this concern.

rudrakshkarpe:
Thanks for catching this. I replaced hasKey with index and added checks so the script stops if either label snapshot fails. I also pinned the RBAC manifest to a commit and added SHA-256 verification before applying it. Both changes are included in the Chinese page too. The full lint, test, and site build pass. @mesutoezdil, could you please take another look for LGTM?

mesutoezdil:
/lgtm

both verified, not just read.

`hasKey` is gone from the english and chinese pages, and i compiled the replacement template with Go `text/template`, it parses. the `|| { echo ...; exit 1; }` guards on both label reads are a nice addition, i had not asked for those.

for the rbac manifest i fetched it at the pinned commit and hashed it:

```
expected e0a77f99422230ccc8958aac0d04694347769279ec26a9d4a5ff729f89efe3d9
actual   e0a77f99422230ccc8958aac0d04694347769279ec26a9d4a5ff729f89efe3d9
```

so the pin and the checksum are both real and a reader running that block gets what you validated.

thanks for the quick turnaround, and sorry for the earlier lgtm before the bot caught line 156.

rudrakshkarpe:
> /lgtm
> 
> both verified, not just read.
> 
> `hasKey` is gone from the english and chinese pages, and i compiled the replacement template with Go `text/template`, it parses. the `|| { echo ...; exit 1; }` guards on both label reads are a nice addition, i had not asked for those.
> 
> for the rbac manifest i fetched it at the pinned commit and hashed it:
> 
> ```
> expected e0a77f99422230ccc8958aac0d04694347769279ec26a9d4a5ff729f89efe3d9
> actual   e0a77f99422230ccc8958aac0d04694347769279ec26a9d4a5ff729f89efe3d9
> ```
> 
> so the pin and the checksum are both real and a reader running that block gets what you validated.
> 
> thanks for the quick turnaround, and sorry for the earlier lgtm before the bot caught line 156.

Thank for the quick review on this, exited to see this lab on the HAMi website soon!

rootsongjc (approved):
/lgtm

PR #655, opened 2026-07-23
docs(tutorials): add Lab 17: Serve Models from a KitOps ModelKit on HAMi

rudrakshkarpe:
## What type of PR is this?

/kind documentation

## What this PR does

Adds **Lab 17: Serve Models from a KitOps ModelKit on HAMi**, an advanced tutorial that replaces a runtime Hugging Face download with a digest-pinned ModelKit pulled from Jozu Hub.

The lab covers:

- inspecting a real, public, digest-pinned Qwen ModelKit
- building checksum-verified amd64 and arm64 KitOps unpacker images
- validating unpack paths and rejecting ambiguous ModelKit layouts
- recovering stale shared-PVC locks with an owner heartbeat
- publishing immutable reference-specific model directories through an atomic symlink switch
- serving the local model with pinned SGLang and vLLM images on HAMi resources
- reserving ephemeral storage and using explicit workload requests
- equivalent English and Chinese instructions
- repacking and pushing a custom ModelKit after Jozu Hub login

## Reference ModelKit

The example uses an existing Qwen ModelKit published by a Jozu Hub user because it is runnable and was verified for this reference lab. It is pinned by digest, so the selected artifact is immutable. The customization section uses `<your-org>` placeholders.

If maintainers prefer project ownership for long-term availability, the same artifact can be mirrored into a HAMi-owned Jozu namespace and the digest reference can be updated.

## Related issue

Addresses #561

## Merge order

SGLang is Lab 15, #793 is Lab 16, and this KitOps tutorial is Lab 17.

## Validation

- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm test` with 93 tests passing
- [x] English and Chinese production build
- [x] Chinese translation added
- [x] Embedded shell syntax checked in both languages
- [x] Path traversal rejection tested
- [x] Ambiguous `config.json` rejection tested
- [x] Stale-lock recovery and symlink publication tested in a container
- [x] KitOps v1.11.0 checksum verification tested
- [x] amd64 and arm64 KitOps image builds tested
- [x] Commits signed off

Verified environment: kind with an NVIDIA H100 80GB cluster. Original workload output was captured on 2026-07-23.

## Summary by CodeRabbit

- **New Features**
  - Added an advanced Lab 17 tutorial covering KitOps ModelKit packaging and model serving on HAMi with SGLang and optional vLLM.
  - Added guidance for secure model unpacking, checksum validation, versioned publishing, shared storage, resource allocation, and recovery workflows.

- **Documentation**
  - Added the tutorial to the Labs sidebar with an estimated duration of about 60 minutes.
  - Added authentication, verification, troubleshooting, cleanup, and cross-node storage guidance.

### DIFF
--- a/sidebars-tutorials.js
+++ b/sidebars-tutorials.js
@@ -61,6 +61,11 @@ module.exports = {
           id: "labs/topology-aware-scheduling",
           customProps: { level: "Intermediate", duration: "about 45 minutes" },
         },
+        {
+          type: "doc",
+          id: "labs/hami-kitops",
+          customProps: { level: "Intermediate", duration: "about 60 minutes" },
+        },
       ],
     },
   ],
--- a/sidebars-tutorials.js
+++ b/sidebars-tutorials.js
@@ -64,7 +64,7 @@ module.exports = {
         {
           type: "doc",
           id: "labs/hami-kitops",
-          customProps: { level: "Intermediate", duration: "about 60 minutes" },
+          customProps: { level: "Advanced", duration: "about 60 minutes" },
         },
       ],
     },
--- a/sidebars-tutorials.js
+++ b/sidebars-tutorials.js
@@ -83,6 +83,11 @@ module.exports = {
         },
         {
           type: "doc",
+          id: "labs/hami-sglang",
+          customProps: { level: "Intermediate", duration: "about 45 minutes" },
+        },
+        {
+          type: "doc",
           id: "labs/hami-kitops",
           customProps: { level: "Advanced", duration: "about 60 minutes" },
         },
--- a/sidebars-tutorials.js
+++ b/sidebars-tutorials.js
@@ -84,7 +84,7 @@ module.exports = {
         {
           type: "doc",
           id: "labs/hami-kitops",
-          customProps: { level: "Intermediate", duration: "about 60 minutes" },
+          customProps: { level: "Advanced", duration: "about 60 minutes" },
         },
       ],
     },
--- a/sidebars-tutorials.js
+++ b/sidebars-tutorials.js
@@ -81,6 +81,11 @@ module.exports = {
           id: "labs/composable-scheduler-policies-gke",
           customProps: { level: "Intermediate", duration: "about 60 minutes" },
         },
+        {
+          type: "doc",
+          id: "labs/hami-kitops",
+          customProps: { level: "Intermediate", duration: "about 60 minutes" },
+        },
       ],
     },
   ],
--- a/tutorials/labs/hami-kitops.md
+++ b/tutorials/labs/hami-kitops.md
@@ -21,7 +21,7 @@ toc_max_heading_level: 2
 
 This lab demonstrates how to replace a runtime **Hugging Face** model download with a **[KitOps](https://kitops.org/) ModelKit** pulled from an OCI registry ([Jozu Hub](https://jozu.ml) in the examples). The model is delivered into the Pod by a KitOps `initContainer`, then served from a **local directory** by [SGLang](https://github.com/sgl-project/sglang) (primary) or vLLM (optional co-resident example) on HAMi-virtualized GPU shares.
 
-Compared with [Lab 6 (vLLM)](./hami-vllm) and [Lab 11 (SGLang)](./hami-sglang), the inference engines still run on HAMi resources — but the **model supply chain** changes from `engine serve <hf-repo>` to **unpack ModelKit → serve local path**.
+Compared with [Lab 6 (vLLM)](./hami-vllm) and Lab 11 (SGLang), the inference engines still run on HAMi resources — but the **model supply chain** changes from `engine serve <hf-repo>` to **unpack ModelKit → serve local path**.
 
 ## Learning Objectives
 
@@ -69,7 +69,7 @@ flowchart TB
 
 ## Prerequisites
 
-- Everything from [Lab 11](./hami-sglang) (or Lab 6): a Kubernetes cluster with NVIDIA GPUs, HAMi installed and healthy, `kubectl`, `helm`
+- Everything from Lab 11 (or Lab 6): a Kubernetes cluster with NVIDIA GPUs, HAMi installed and healthy, `kubectl`, `helm`
 - Docker (or an equivalent builder) to build and load images into the cluster
 - [`kit`](https://github.com/jozu-ai/kitops) CLI on your workstation (optional but recommended for `kit inspect`)
 - Ability to pull from the public registry `jozu.ml` (no login required for the sample ModelKit)
@@ -776,4 +776,4 @@ kubectl delete namespace kitops --ignore-not-found
 - Swap the public Jozu ModelKit for your internal registry ModelKit and wire imagePullSecrets / `kit login` Secrets.
 - Share one PVC across SGLang and vLLM so the ModelKit is unpacked once.
 - Combine with [Lab 3: GPU Partitioning](./gpu-partitioning) to pack more tenants per GPU.
-- Return to [Lab 11: SGLang](./hami-sglang) if you need the simpler Hugging Face runtime path for debugging engines independently of the supply chain.
+- Return to Lab 11: SGLang if you need the simpler Hugging Face runtime path for debugging engines independently of the supply chain.
--- a/tutorials/labs/hami-kitops.md
+++ b/tutorials/labs/hami-kitops.md
@@ -1,6 +1,6 @@
 ---
 title: "Lab 12: Serve Models from a KitOps ModelKit on HAMi"
-description: "Replace runtime Hugging Face downloads with a KitOps ModelKit delivered by an initContainer, then serve locally with SGLang (and optionally vLLM) on HAMi GPU shares."
+description: "Package a model as a KitOps ModelKit, pull it from Jozu Hub with an initContainer, and serve it locally with SGLang (and optionally vLLM) on HAMi GPU shares."
 sidebar_label: "Lab 12: KitOps ModelKit Inference"
 lab:
   level: Intermediate
@@ -19,15 +19,15 @@ tags:
 toc_max_heading_level: 2
 ---
 
-This lab demonstrates how to replace a runtime **Hugging Face** model download with a **[KitOps](https://kitops.org/) ModelKit** pulled from an OCI registry ([Jozu Hub](https://jozu.ml) in the examples). The model is delivered into the Pod by a KitOps `initContainer`, then served from a **local directory** by [SGLang](https://github.com/sgl-project/sglang) (primary) or vLLM (optional co-resident example) on HAMi-virtualized GPU shares.
+This lab demonstrates how to package a model as a **[KitOps](https://kitops.org/) ModelKit** — a versioned OCI artifact — and download it from an OCI registry (**[Jozu Hub](https://jozu.ml)** in the examples) into the Pod using a KitOps `initContainer`, then serve it from a **local directory** with [SGLang](https://github.com/sgl-project/sglang) (primary) or vLLM (optional co-resident example) on HAMi-virtualized GPU shares.
 
-Compared with [Lab 6 (vLLM)](./hami-vllm) and Lab 11 (SGLang), the inference engines still run on HAMi resources — but the **model supply chain** changes from `engine serve <hf-repo>` to **unpack ModelKit → serve local path**.
+Like [Lab 6 (vLLM)](./hami-vllm) and Lab 11 (SGLang), the inference engines run on HAMi resources. Here the **model supply chain** is registry-native: the model is packaged as a ModelKit, versioned and stored on Jozu Hub, pulled into the Pod as an OCI artifact, and served from a local path.
 
 ## Learning Objectives
 
 - Inspect a public KitOps ModelKit on an OCI registry
 - Build a small `kitunpacker` init image and a custom SGLang serve image
-- Deploy a Pod that pulls/unpacks a ModelKit via `initContainer` (no HF download in the main container)
+- Deploy a Pod that pulls/unpacks a ModelKit via `initContainer` (the main container loads the model from the KitOps-delivered volume)
 - Schedule the workload with HAMi `nvidia.com/gpu` / `gpumem` / `gpucores`
 - Prove inference works against the OpenAI-compatible SGLang API
 - Optionally co-locate a vLLM Pod serving the same ModelKit on the same physical GPU
@@ -128,7 +128,7 @@ Example (truncated) output from the verification environment:
 }
 ```
 
-The ModelKit carries safetensors weights plus tokenizer/config as OCI layers. The initContainer will unpack and flatten them into a Hugging Face-style directory that SGLang/vLLM can load locally.
+The ModelKit carries safetensors weights plus tokenizer/config as OCI layers. The initContainer will unpack and flatten them into a flat model directory (`config.json` + `*.safetensors`) that SGLang/vLLM can load locally.
 
 ## Step 3: Build the Pipeline Images
 
@@ -162,8 +162,8 @@ ENTRYPOINT ["/usr/local/bin/unpack.sh"]
 ```sh
 #!/usr/bin/env sh
 # kitunpacker: pull a ModelKit from an OCI registry (Jozu Hub by default) and
-# unpack the model into a Hugging Face-style directory that vLLM / SGLang can
-# load directly -- no Hugging Face download at runtime.
+# unpack the model into a flat directory (config.json + *.safetensors) that
+# vLLM / SGLang can load directly from local disk.
 #
 # Env (all overridable from the Pod spec):
 #   MODELKIT_REF   full ModelKit reference, e.g. jozu.ml/<org>/<repo>:<tag>
@@ -244,7 +244,7 @@ Build:
 docker build -t hami-kitunpacker:latest ./kitunpacker
 ```
 
-### 3.2 Custom SGLang image (local model path, no HF repo)
+### 3.2 Custom SGLang image (serves a local model path)
 
 `sglang/Dockerfile`:
 
@@ -270,7 +270,7 @@ ENTRYPOINT ["/usr/local/bin/serve.sh"]
 #!/usr/bin/env bash
 # Custom SGLang entrypoint: serve a model unpacked from a KitOps ModelKit that
 # the kitunpacker initContainer placed on a shared volume. Serves from a LOCAL
-# directory (--model-path) so there is no Hugging Face download at runtime.
+# directory (--model-path) delivered straight from the Jozu Hub ModelKit.
 set -euo pipefail
 
 MODEL_DIR="${MODEL_DIR:-/models/qwen3}"
@@ -280,7 +280,7 @@ MEM_FRACTION="${MEM_FRACTION:-0.8}"
 PORT="${PORT:-30000}"
 ATTENTION_BACKEND="${ATTENTION_BACKEND:-triton}"
 
-echo "[sglang-jozu] serving KitOps model from ${MODEL_DIR} (source: Jozu Hub, no HF download)"
+echo "[sglang-jozu] serving KitOps model from ${MODEL_DIR} (source: Jozu Hub ModelKit)"
 if [ ! -f "${MODEL_DIR}/config.json" ]; then
   echo "[sglang-jozu] ERROR: ${MODEL_DIR}/config.json not found -- did the kitunpacker init run?" >&2
   exit 1
@@ -338,9 +338,8 @@ ENTRYPOINT ["/usr/local/bin/serve.sh"]
 ```bash
 #!/usr/bin/env bash
 # Custom vLLM entrypoint: serve a model unpacked from a KitOps ModelKit that the
-# kitunpacker initContainer placed on a shared volume. This deliberately serves
-# from a LOCAL directory instead of the default `vllm serve <hf-repo>` so there
-# is no Hugging Face download at runtime.
+# kitunpacker initContainer placed on a shared volume. It serves from a LOCAL
+# directory (--model-path) populated from the Jozu Hub ModelKit.
 set -euo pipefail
 
 MODEL_DIR="${MODEL_DIR:-/models/qwen3}"
@@ -349,7 +348,7 @@ MAX_MODEL_LEN="${MAX_MODEL_LEN:-8192}"
 GPU_MEM_UTIL="${GPU_MEM_UTIL:-0.85}"
 PORT="${PORT:-8000}"
 
-echo "[vllm-jozu] serving KitOps model from ${MODEL_DIR} (source: Jozu Hub, no HF download)"
+echo "[vllm-jozu] serving KitOps model from ${MODEL_DIR} (source: Jozu Hub ModelKit)"
 if [ ! -f "${MODEL_DIR}/config.json" ]; then
   echo "[vllm-jozu] ERROR: ${MODEL_DIR}/config.json not found -- did the kitunpacker init run?" >&2
   exit 1
@@ -515,10 +514,10 @@ kubectl -n kitops rollout status deploy/sglang-modelkit --timeout=30m
 kubectl -n kitops logs -l app.kubernetes.io/name=sglang-modelkit -c sglang --tail=50
 ```
 
-You should see the custom entrypoint message proving a **local** ModelKit path (not an HF repo id):
+You should see the custom entrypoint message confirming the model is served from the **local** ModelKit path:
 
 ```plaintext
-[sglang-jozu] serving KitOps model from /models/qwen3 (source: Jozu Hub, no HF download)
+[sglang-jozu] serving KitOps model from /models/qwen3 (source: Jozu Hub ModelKit)
 ... model_path='/models/qwen3' ... served_model_name='qwen3-4b-instruct' ...
 ```
 
@@ -706,7 +705,7 @@ Ensure combined `gpumem` requests fit on the physical GPU (for example two × 30
 #     jozu.ml/jonathangamer202002/qwen3-4b-instruct:latest
 #
 # This Kitfile is provided so you can (re)pack and push your OWN ModelKit to a
-# registry (Jozu Hub, ACR, GHCR, ...) from a local HF-format model directory:
+# registry (Jozu Hub, ACR, GHCR, ...) from a local model directory (config.json + safetensors):
 #
 #     # 1) get a model directory (e.g. via `kit unpack` or `kit import`)
 #     # 2) place this Kitfile next to a ./qwen3 directory of safetensors + config
@@ -720,7 +719,7 @@ package:
   authors:
     - HAMi KubeCon Demo
   description: >
-    Qwen3-4B-Instruct-2507 packaged as a KitOps ModelKit (HF safetensors layout),
+    Qwen3-4B-Instruct-2507 packaged as a KitOps ModelKit (safetensors layout),
     served on HAMi-virtualized GPUs by vLLM and SGLang.
 model:
   name: qwen3-4b-instruct
@@ -734,7 +733,7 @@ docs:
 ```
 
 ```bash
-# After placing a HF-layout model directory at ./qwen3 next to the Kitfile:
+# After placing a safetensors-layout model directory at ./qwen3 next to the Kitfile:
 kit pack . -t jozu.ml/<your-org>/qwen3-4b-instruct:latest
 kit push jozu.ml/<your-org>/qwen3-4b-instruct:latest
 ```
@@ -766,7 +765,7 @@ kubectl delete namespace kitops --ignore-not-found
 | Claim | Evidence |
 | --- | --- |
 | Model is an OCI ModelKit | `kit inspect --remote` returns KitOps manifest / model layers. |
-| Model delivered without HF in the main container | initContainer logs show `kit unpack`; SGLang logs show `serving KitOps model from /models/qwen3` and `model_path='/models/qwen3'`. |
+| Model delivered from the ModelKit into the main container | initContainer logs show `kit unpack`; SGLang logs show `serving KitOps model from /models/qwen3` and `model_path='/models/qwen3'`. |
 | HAMi schedules the workload | `schedulerName: hami-scheduler` + Filtering/Binding events. |
 | GPU memory/compute caps apply | `CUDA_DEVICE_MEMORY_LIMIT_0=30000m`, `CUDA_DEVICE_SM_LIMIT=30`; in-pod `nvidia-smi` shows `... / 30000MiB`. |
 | Inference works | `/v1/models` lists `qwen3-4b-instruct`; chat completions return content. |
@@ -776,4 +775,4 @@ kubectl delete namespace kitops --ignore-not-found
 - Swap the public Jozu ModelKit for your internal registry ModelKit and wire imagePullSecrets / `kit login` Secrets.
 - Share one PVC across SGLang and vLLM so the ModelKit is unpacked once.
 - Combine with [Lab 3: GPU Partitioning](./gpu-partitioning) to pack more tenants per GPU.
-- Return to Lab 11: SGLang if you need the simpler Hugging Face runtime path for debugging engines independently of the supply chain.
+- Return to Lab 11: SGLang for the simpler path where the engine pulls the model directly at startup, useful for debugging engines independently of the supply chain.
--- a/tutorials/labs/hami-kitops.md
+++ b/tutorials/labs/hami-kitops.md
@@ -1,0 +1,779 @@
+---
+title: "Lab 12: Serve Models from a KitOps ModelKit on HAMi"
+description: "Replace runtime Hugging Face downloads with a KitOps ModelKit delivered by an initContainer, then serve locally with SGLang (and optionally vLLM) on HAMi GPU shares."
+sidebar_label: "Lab 12: KitOps ModelKit Inference"
+lab:
+  level: Intermediate
+  duration: about 60 minutes
+  environment: Kubernetes cluster with NVIDIA GPUs
+  authors:
+    - rudrakshkarpe
+    - shivaylamba
+  verified: "2026-07-23"
+tags:
+  - kitops
+  - sglang
+  - vllm
+  - inference
+  - modelkit
+toc_max_heading_level: 2
+---
+
+This lab demonstrates how to replace a runtime **Hugging Face** model download with a **[KitOps](https://kitops.org/) ModelKit** pulled from an OCI registry ([Jozu Hub](https://jozu.ml) in the examples). The model is delivered into the Pod by a KitOps `initContainer`, then served from a **local directory** by [SGLang](https://github.com/sgl-project/sglang) (primary) or vLLM (optional co-resident example) on HAMi-virtualized GPU shares.
+
+Compared with [Lab 6 (vLLM)](./hami-vllm) and [Lab 11 (SGLang)](./hami-sglang), the inference engines still run on HAMi resources — but the **model supply chain** changes from `engine serve <hf-repo>` to **unpack ModelKit → serve local path**.
+
+## Learning Objectives
+
+- Inspect a public KitOps ModelKit on an OCI registry
+- Build a small `kitunpacker` init image and a custom SGLang serve image
+- Deploy a Pod that pulls/unpacks a ModelKit via `initContainer` (no HF download in the main container)
+- Schedule the workload with HAMi `nvidia.com/gpu` / `gpumem` / `gpucores`
+- Prove inference works against the OpenAI-compatible SGLang API
+- Optionally co-locate a vLLM Pod serving the same ModelKit on the same physical GPU
+
+## Lab Overview
+
+```mermaid
+%% title: HAMi + KitOps ModelKit Lab Flowchart
+flowchart LR
+    Step1["Step 1<br/>Prereqs + HAMi"] --> Step2["Step 2<br/>Inspect ModelKit"]
+    Step2 --> Step3["Step 3<br/>Build Images"]
+    Step3 --> Step4["Step 4<br/>Deploy SGLang+ModelKit"]
+    Step4 --> Step5["Step 5<br/>Watch Unpack"]
+    Step5 --> Step6["Step 6<br/>Test Inference"]
+    Step6 --> Step7["Step 7<br/>Verify HAMi Caps"]
+    Step7 --> Step8["Step 8<br/>Optional vLLM"]
+    Step8 --> Step9["Step 9<br/>Cleanup"]
+```
+
+## Deployment Architecture
+
+```mermaid
+%% title: KitOps ModelKit delivery on HAMi
+flowchart TB
+    Registry["OCI Registry<br/>Jozu Hub ModelKit"] -->|"kit unpack --filter model"| Init["initContainer<br/>hami-kitunpacker"]
+    Init --> Vol["emptyDir /models"]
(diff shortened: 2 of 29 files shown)


mesutoezdil (commented):
solid lab, the initcontainer + flatten trick is neat and i checked the `kit inspect` digest against the live registry, it matches (`sha256:df4629f6...`) so the captured output is real. two things before this is mergeable: the `format:check` box is ticked but `npx prettier --check tutorials/labs/hami-kitops.md` fails on the kitfile block, and there is no zh mirror while the other 10 labs all have one under `i18n/zh/docusaurus-plugin-content-docs-tutorials/current/labs/`. rest is inline. build passes locally and no broken links.

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,778 @@
+---
+title: "Lab 12: Serve Models from a KitOps ModelKit on HAMi"
+description: "Package a model as a KitOps ModelKit, pull it from Jozu Hub with an initContainer, and serve it locally with SGLang (and optionally vLLM) on HAMi GPU shares."
+sidebar_label: "Lab 12: KitOps ModelKit Inference"
+lab:
+  level: Intermediate
```
mesutoezdil: this asks the reader to write 4 files and build 3 images, and the other 60 min labs (volcano, kueue) are Advanced, so imo this is Advanced too

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,778 @@
...
+This lab assumes HAMi is already installed as in Lab 11. If not, complete Lab 11 Steps 1–3 first.
+
+## Example Cluster State
+
+Verification used the same kind + H100 cluster as Lab 11, with HAMi advertising 10 vGPUs and both `hami-scheduler` / `hami-device-plugin` Running.
+
+Public ModelKit used throughout:
+
+```plaintext
+jozu.ml/jonathangamer202002/qwen3-4b-instruct:latest
```
mesutoezdil: the whole lab hangs off one personal account and a mutable tag, so pls pin by digest and ideally mirror the modelkit under a project owned repo, otherwise the lab breaks the day that account is cleaned up

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,778 @@
...
+kubectl get pods -n kube-system -l app.kubernetes.io/instance=hami -o wide
+kubectl get nodes -o 'custom-columns=NAME:.metadata.name,GPU:.status.allocatable.nvidia\.com/gpu'
+```
+
+Expected: device plugin and scheduler Running; GPU nodes show allocatable `nvidia.com/gpu` (for example `10`).
+
+Scale down any large competing GPU workloads so the ModelKit + 4B model can fit:
+
+```bash
+kubectl -n sglang scale deploy/sglang-qwen3-17b --replicas=0 2>/dev/null || true
```
mesutoezdil: this scales a deploy that only exists if the reader did lab 11, and `|| true` hides that, so a generic "free up gpu shares" line would age better?

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,778 @@
...
+
+### 3.1 `kitunpacker` init image
+
+`kitunpacker/Dockerfile`:
+
+```dockerfile
+FROM alpine:3.20
+
+RUN apk add --no-cache bash coreutils findutils ca-certificates curl tar \
+    && curl -fsSL "https://github.com/jozu-ai/kitops/releases/latest/download/kitops-linux-x86_64.tar.gz" -o /tmp/kit.tgz \
```
mesutoezdil: base image is pinned to alpine 3.20 but the kit cli is `releases/latest`, so pin the kitops version too or the init image quietly changes between builds?

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,778 @@
...
+echo "[kitunpacker] ref=${MODELKIT_REF} -> ${DEST}"
+
+if ready; then
+  echo "[kitunpacker] model already present, skipping unpack"
+  exit 0
+fi
+
+# simple cross-pod lock (mkdir is atomic) so two engines sharing one cache
+# don't race to write the same files.
+if ! mkdir "${LOCK}" 2>/dev/null; then
```
mesutoezdil, line 200: with `emptyDir` each pod gets its own volume so this lock never sees a peer, as step 8 says itself, worth a note that it only matters on a shared pvc

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,778 @@
...
+```bash
+docker build -t hami-kitunpacker:latest ./kitunpacker
+```
+
+### 3.2 Custom SGLang image (serves a local model path)
+
+`sglang/Dockerfile`:
+
+```dockerfile
+FROM lmsysorg/sglang:latest
```
mesutoezdil: `:latest` here and on `vllm/vllm-openai` at line 322 would be the only ones under `tutorials/`, and it clashes with the `verified: 2026-07-23` claim since the base can change under you

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,778 @@
...
+            failureThreshold: 90
+          volumeMounts:
+            - name: modelkit
+              mountPath: /models
+            - name: dshm
+              mountPath: /dev/shm
+      volumes:
+        - name: modelkit
+          emptyDir:
+            sizeLimit: 20Gi
```
mesutoezdil, line 473: fyi the model layer is ~7.5 GiB and it lands in `KITOPS_HOME` and then in `.raw-qwen3` before the mv, so peak is ~15 GiB, worth saying so nobody trims this to 10Gi?

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,778 @@
...
+#     kit pack . -t jozu.ml/<your-org>/qwen3-4b-instruct:latest
+#     kit login jozu.ml -u <user> --password-stdin   # needed only for push
+#     kit push jozu.ml/<your-org>/qwen3-4b-instruct:latest
+manifestVersion: "1.0"
+package:
+  name: qwen3-4b-instruct
+  version: "1.0"
+  authors:
+    - HAMi KubeCon Demo
+  description: >
```
mesutoezdil, line 727: `prettier --check` fails only on this yaml block (folded description + trailing blank line), so `npm run format` fixes it and unblocks docs-health ci..

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,778 @@
...
+    Qwen3-4B-Instruct-2507 packaged as a KitOps ModelKit (safetensors layout),
+    served on HAMi-virtualized GPUs by vLLM and SGLang.
+model:
+  name: qwen3-4b-instruct
+  path: ./qwen3
+  license: Apache-2.0
+  description: Qwen3 4B instruct, safetensors (Qwen3ForCausalLM)
+docs:
+  - path: ./README.md
```
mesutoezdil: kit resolves kitfile paths against the pack context and the steps only tell you to place `./qwen3`, so `kit pack .` trips on the missing `./README.md`?

sidebars-tutorials.js
```diff
@@ -61,6 +61,11 @@ module.exports = {
           id: "labs/topology-aware-scheduling",
           customProps: { level: "Intermediate", duration: "about 45 minutes" },
         },
+        {
+          type: "doc",
+          id: "labs/hami-kitops",
```
mesutoezdil, line 66: #654 appends to this same list, so whichever lands second needs a rebase and the "lab 12" title only holds if #654 goes first.

tutorials/labs/hami-kitops.md
rudrakshkarpe: That makes sense given the amount of setup involved. I have moved it to Advanced in both the frontmatter and sidebar.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Thanks for checking the digest. I have pinned the ModelKit to the verified digest so the lab no longer depends on `latest`. Is there a preferred HAMi-owned namespace on Jozu Hub that we should use for the mirror? I would be happy to update the reference once that location is confirmed.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Yes, the Lab 11 deployment was too specific here. I replaced it with general guidance for checking current GPU usage and freeing enough HAMi capacity.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Nice catch. KitOps is now pinned to v1.11.0, so rebuilding the init image will not silently pull a different release.

tutorials/labs/hami-kitops.md
rudrakshkarpe: You are right. I clarified that this lock only coordinates Pods using the same shared PVC. With the default `emptyDir`, every Pod has its own volume and lock.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Makes sense. I pinned SGLang to v0.5.7 and vLLM to v0.23.0 so the documented setup stays reproducible.

tutorials/labs/hami-kitops.md
rudrakshkarpe: That is helpful context. I added the roughly 15 GiB peak unpack usage and explained why the example keeps the volume at 20 GiB or larger.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Thanks for flagging this. I ran the repository formatter and confirmed that `npm run format:check` now passes.

tutorials/labs/hami-kitops.md
rudrakshkarpe: You are right, the example did not create that file. I removed the unnecessary `docs` entry, so the shown `kit pack .` flow is now self-contained.

sidebars-tutorials.js
rudrakshkarpe: Yes, this should follow #654. I will rebase it after #654 lands and make sure the sidebar order and Lab 12 numbering remain correct.

rudrakshkarpe:
Thanks for going through the lab so carefully, and for validating the ModelKit digest. I have worked through the inline notes, added the Chinese version, and rerun the formatting and build checks. This PR will stay behind #654, and I will rebase it once that PR lands. I also pinned the current ModelKit by digest. If there is a preferred HAMi-owned Jozu Hub namespace, I would be glad to move the example there.

mesutoezdil (commented):
solid lab. rebase on master, numbering has moved.

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,781 @@
+---
+title: "Lab 12: Serve Models from a KitOps ModelKit on HAMi"
```
mesutoezdil, line 2: lab 12 is taken on master. renumber to lab 15, also in sidebar_label, pr title and the zh file.

sidebars-tutorials.js
mesutoezdil, line 66: move last, after labs/volcano-ascend-vnpu.

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,781 @@
...
+  - sglang
+  - vllm
+  - inference
+  - modelkit
+toc_max_heading_level: 2
+---
+
+This lab demonstrates how to package a model as a **[KitOps](https://kitops.org/) ModelKit**, a versioned OCI artifact, and download it from an OCI registry (**[Jozu Hub](https://jozu.ml)** in the examples) into the Pod using a KitOps `initContainer`, then serve it from a **local directory** with [SGLang](https://github.com/sgl-project/sglang) (primary) or vLLM (optional co-resident example) on HAMi-virtualized GPU shares.
+
+Like [Lab 6 (vLLM)](./hami-vllm) and Lab 11 (SGLang), the inference engines run on HAMi resources. Here the **model supply chain** is registry-native: the model is packaged as a ModelKit, versioned and stored on Jozu Hub, pulled into the Pod as an OCI artifact, and served from a local path.
```
mesutoezdil, line 24: the lab 11 references in prose need the new number too.

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,781 @@
...
+This lab assumes HAMi is already installed as in Lab 11. If not, complete Lab 11 Steps 1–3 first.
+
+## Example Cluster State
+
+Verification used the same kind + H100 cluster as Lab 11, with HAMi advertising 10 vGPUs and both `hami-scheduler` / `hami-device-plugin` Running.
+
+Public ModelKit used throughout:
+
+```plaintext
+jozu.ml/jonathangamer202002/qwen3-4b-instruct@sha256:df4629f6a10bba7bec45e12bd15f910ed1024699bfbb44b63240899f71bb1c19
```
mesutoezdil, line 86: personal jozu namespace. a project-owned mirror would be safer, as you offered.

sidebars-tutorials.js
rudrakshkarpe: Resolved during the rebase: the KitOps entry is now last in the current sidebar, after the newer Lab 14 (`composable-scheduler-policies-gke`).

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,781 @@
...
+```dockerfile
+FROM alpine:3.20
+
+ARG KITOPS_VERSION=v1.11.0
+
+RUN apk add --no-cache bash coreutils findutils ca-certificates curl tar \
+    && curl -fsSL "https://github.com/kitops-ml/kitops/releases/download/${KITOPS_VERSION}/kitops-linux-x86_64.tar.gz" -o /tmp/kit.tgz \
+    && tar -xzf /tmp/kit.tgz -C /usr/local/bin kit \
+    && rm -f /tmp/kit.tgz \
+    && kit version
```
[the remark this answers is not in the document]
rudrakshkarpe: Fixed in a6bb97b0. Both Dockerfile snippets now download the official v1.11.0 checksum manifest and verify the Linux x86_64 archive with `sha256sum -c` before extraction. I also verified the documented pipeline locally against the release asset.

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,781 @@
...
+[ -n "${SRC_CFG}" ] || { echo "[kitunpacker] config.json not found after unpack" >&2; exit 1; }
+SRC="$(dirname "${SRC_CFG}")"
+
+mkdir -p "${DEST}"
+# all weight shards, wherever they live under the unpacked tree
+find "${SRC}" -name '*.safetensors' -exec mv -f {} "${DEST}/" \;
+# all top-level metadata files (config, index, tokenizer, vocab, generation cfg)
+find "${SRC}" -maxdepth 1 -type f -exec mv -f {} "${DEST}/" \;
+
+rm -rf "${RAW}" "${KITOPS_HOME}"
```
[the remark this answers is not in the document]
rudrakshkarpe: Fixed in a6bb97b0 in both languages. An `EXIT` cleanup trap is installed before optional login and removes `RAW`, `STAGE`, `KITOPS_HOME`, and the lock. Explicit INT/TERM traps preserve non-zero signal exits, so credentials/cache are cleaned when login or unpack fails.

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,781 @@
...
+fi
+
+# This lock only coordinates Pods when they mount the same shared PVC. With
+# the emptyDir used in this lab, every Pod has an isolated volume and lock.
+# mkdir is atomic, so Pods sharing a PVC do not race to write the same files.
+if ! mkdir "${LOCK}" 2>/dev/null; then
+  echo "[kitunpacker] another unpack in progress, waiting for it to finish..."
+  i=0
+  while [ "${i}" -lt 360 ]; do
+    ready && { echo "[kitunpacker] model became ready"; exit 0; }
```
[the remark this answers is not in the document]
rudrakshkarpe: Fixed in a6bb97b0 in both languages. `ready()` now requires a validated model plus a `.modelkit-ref` marker whose contents exactly match `MODELKIT_REF`. Unpack/flatten happens in a staging directory, validation and marker creation complete there, and only then is the staged tree renamed into `DEST`.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Rebased onto current `master` and renumbered the English and Chinese title/sidebar labels to Lab 15 in a6bb97b0. I also updated the PR title.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Fixed in a6bb97b0. All stale Lab 11/SGLang references were removed in both languages. Setup now links to the stable Lab 6 HAMi/vLLM prerequisites, and verification/troubleshooting text is self-contained.

tutorials/labs/hami-kitops.md
rudrakshkarpe: This ModelKit was selected because it is an existing, working Qwen artifact published by a Jozu Hub user and verified for this reference lab. The example is pinned by digest, so the exact artifact used by the lab is immutable and reproducible.

My preference is to keep a concrete, tested reference here rather than replace it entirely with a templated placeholder, since readers can run the lab as written. We can still use placeholders such as `jozu.ml/<your-org>/...` in the repack/customization section.

I agree that project ownership would improve long-term availability. If the maintainers prefer, we can mirror the same ModelKit into a HAMi-owned Jozu namespace and update `MODELKIT_REF` to that digest. Please let me know the preferred namespace—or whether keeping this verified digest is acceptable for the reference example—and I will align the lab accordingly.

rudrakshkarpe:
All actionable review feedback is addressed after rebasing onto current `master`:

- coordinated numbering with companion SGLang PR #654: SGLang is Lab 15 and this dependent KitOps tutorial is Lab 16
- placed the sidebar entry after the current labs; it will follow SGLang in the intended merge order
- removed stale Lab 11/SGLang references and linked stable HAMi setup guidance
- verified the KitOps archive against the official v1.11.0 checksum before extraction
- made ModelKit readiness exact-reference-aware and publication staging-based
- added failure/signal-safe cleanup for unpack data, cache, credentials, and lock
- kept the English and Chinese snippets in sync

Validation completed:
- `npm run lint`
- `npm test` (73/73)
- Prettier checks for both lab files and sidebar
- POSIX `sh -n` for both embedded unpacker scripts
- live checksum verification against the KitOps v1.11.0 release assets
- production build output generated for both `en` and `zh`
- signed-off commits with DCO passing

The remaining discussion is about the reference ModelKit location: the current user-published Qwen ModelKit is a real, verified artifact and is pinned by digest for reproducibility. I recommend retaining a runnable concrete example while keeping `<your-org>` placeholders in the customization section. If project ownership is preferred for long-term availability, I am also happy to switch the lab to the same artifact mirrored under a HAMi-owned Jozu namespace once the preferred namespace is provided.

This PR should merge after #654 so the Lab 15 → Lab 16 ordering remains contiguous.

rudrakshkarpe:
Numbering coordination update: #654 is now Lab 15 on current `master`, so this dependent KitOps lab is now Lab 16 in English, Chinese, and the PR title (0b5955f8). This preserves the intended SGLang → KitOps learning order; please merge #654 before #655.

mesutoezdil:
Closing this for now: the tutorial section is one of the areas #689 is reorganizing, so new labs are on hold until the new structure lands.

mesutoezdil:
reopening: #654 is back open and links to this lab, so the two should be reviewed together.

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,805 @@
...
+
+`kitunpacker/Dockerfile`:
+
+```dockerfile
+FROM alpine:3.20
+
+ARG KITOPS_VERSION=v1.11.0
+
+RUN apk add --no-cache bash coreutils findutils ca-certificates curl tar \
+    && curl -fsSL "https://github.com/kitops-ml/kitops/releases/download/${KITOPS_VERSION}/kitops-linux-x86_64.tar.gz" -o /tmp/kit.tgz \
```
mesutoezdil: alpine:3.20 is multi arch, so a build on an apple silicon laptop produces an arm64 image with an x86_64 kit binary, and the `kit version` on the last line of this RUN would fail. should the arch come from TARGETARCH?

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,805 @@
...
+find "${SRC}" -name '*.safetensors' -exec mv -f {} "${STAGE}/" \;
+# all top-level metadata files (config, index, tokenizer, vocab, generation cfg)
+find "${SRC}" -maxdepth 1 -type f -exec mv -f {} "${STAGE}/" \;
+
+valid_model "${STAGE}" || { echo "[kitunpacker] validation failed: missing config or shards" >&2; exit 1; }
+printf '%s\n' "${MODELKIT_REF}" >"${STAGE}/${MARKER}"
+
+# Rename only a fully validated tree into place. Readers never observe a
+# partially populated DEST, and a stale DEST for another reference is replaced.
+rm -rf "${DEST}"
```
mesutoezdil: the comment above says readers never observe a partially populated DEST, but rm followed by mv leaves a window where DEST does not exist at all. is that safe for a peer pod already serving from it?

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,805 @@
...
+fi
+
+rm -rf "${RAW}" "${STAGE}"; mkdir -p "${RAW}" "${STAGE}"
+echo "[kitunpacker] pulling + unpacking model layers from registry..."
+kit unpack --filter model "${MODELKIT_REF}" -d "${RAW}"
+
+# Flatten: ModelKits may store the .safetensors shards in a model/ subdir while
+# config.json / *.index.json / tokenizer sit one level up. vLLM/transformers
+# need them all in one directory, so collect everything into a staging directory.
+SRC_CFG="$(find "${RAW}" -name config.json | head -1)"
```
mesutoezdil: find returns filesystem order, so head -1 is not deterministic if a modelkit ships more than one config.json. is one guaranteed here?

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,805 @@
...
+
+if ready; then
+  echo "[kitunpacker] model already present, skipping unpack"
+  exit 0
+fi
+
+# This lock only coordinates Pods when they mount the same shared PVC. With
+# the emptyDir used in this lab, every Pod has an isolated volume and lock.
+# mkdir is atomic, so Pods sharing a PVC do not race to write the same files.
+if ! mkdir "${LOCK}" 2>/dev/null; then
```
mesutoezdil: on the shared pvc case this comment describes, a pod killed after mkdir leaves the lock forever and every later pod waits 30 minutes and fails. is there a way out without deleting the directory by hand?

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,805 @@
...
+
+    HAMI["HAMi<br/>gpumem + gpucores caps"] --> Pod
+    GPU["Physical NVIDIA GPU"] --> HAMI
+```
+
+## Prerequisites
+
+- A Kubernetes cluster with NVIDIA GPUs, HAMi installed and healthy, `kubectl`, and `helm` (see [Lab 6](./hami-vllm) for a complete setup)
+- Docker (or an equivalent builder) to build and load images into the cluster
+- [`kit`](https://github.com/jozu-ai/kitops) CLI on your workstation (optional but recommended for `kit inspect`)
```
mesutoezdil: this links jozu-ai/kitops while the dockerfile downloads from kitops-ml/kitops. the first one redirects. should both use the current name?

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,805 @@
+---
+title: "Lab 16: Serve Models from a KitOps ModelKit on HAMi"
```
mesutoezdil, line 2: correction on my earlier note: master already has Lab 14, so Lab 16 here is right and there is no gap. the pr body still says Lab 12 though. could that be refreshed?

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,805 @@
...
+  name: qwen3-4b-instruct
+  path: ./qwen3
+  license: Apache-2.0
+  description: Qwen3 4B instruct, safetensors (Qwen3ForCausalLM)
+```
+
+```bash
+# After placing a safetensors-layout model directory at ./qwen3 next to the Kitfile:
+kit pack . -t jozu.ml/<your-org>/qwen3-4b-instruct:latest
+kit push jozu.ml/<your-org>/qwen3-4b-instruct:latest
```
mesutoezdil, line 850: the comment block above lists `kit login` before push, but this runnable block goes straight from pack to push, so it would 401 on a fresh shell. should login be in here too?

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,805 @@
...
+          env:
+            - name: MODEL_DIR
+              value: "/models/qwen3"
+            - name: SERVED_NAME
+              value: "qwen3-4b-instruct"
+          ports:
+            - name: http
+              containerPort: 8000
+          resources:
+            limits:
```
mesutoezdil, line 763: this container sets limits with no requests, so kubernetes copies them and the pod asks for 8 cpu and 32Gi, while the sglang one above asks for 2 and 8Gi. can both land on the single node used for verification?

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,805 @@
...
+            failureThreshold: 90
+          volumeMounts:
+            - name: modelkit
+              mountPath: /models
+            - name: dshm
+              mountPath: /dev/shm
+      volumes:
+        - name: modelkit
+          emptyDir:
+            sizeLimit: 20Gi
```
mesutoezdil, line 571: the troubleshooting table names disk pressure on this emptyDir, but the pod requests no ephemeral-storage, so the scheduler does not reserve the 20Gi. should there be a request for it?

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,805 @@
...
+| Model is an OCI ModelKit | `kit inspect --remote` returns KitOps manifest / model layers. |
+| Model delivered from the ModelKit into the main container | initContainer logs show `kit unpack`; SGLang logs show `serving KitOps model from /models/qwen3` and `model_path='/models/qwen3'`. |
+| HAMi schedules the workload | `schedulerName: hami-scheduler` + Filtering/Binding events. |
+| GPU memory/compute caps apply | `CUDA_DEVICE_MEMORY_LIMIT_0=30000m`, `CUDA_DEVICE_SM_LIMIT=30`; in-pod `nvidia-smi` shows `... / 30000MiB`. |
+| Inference works | `/v1/models` lists `qwen3-4b-instruct`; chat completions return content. |
+
+## Next Steps
+
+- Swap the public Jozu ModelKit for your internal registry ModelKit and wire imagePullSecrets / `kit login` Secrets.
+- Share one PVC across SGLang and vLLM so the ModelKit is unpacked once.
```
mesutoezdil: two pods sharing one pvc needs ReadWriteMany, which the kind default local-path provisioner does not offer. is that worth a note here?

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,805 @@
...
+```
+
+Test with:
+
+```bash
+kubectl -n kitops port-forward svc/vllm-modelkit 8000:8000
+curl -s http://127.0.0.1:8000/v1/models
+```
+
+Ensure combined `gpumem` requests fit on the physical GPU (for example two × 30000 MiB needs ≥ 60 GiB free on that card).
```
mesutoezdil: two times 30000 MiB is 58.6 GiB, not 60. the rest of the lab is careful about MiB, should this line be too?

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,805 @@
...
+#   MODELKIT_REF   full ModelKit reference, e.g. jozu.ml/<org>/<repo>:<tag>
+#   UNPACK_PATH    root volume to unpack into                 (default /models)
+#   MODEL_SUBDIR   final model dir under UNPACK_PATH          (default qwen3)
+#   REGISTRY_URL/USERNAME/PASSWORD  optional creds for PRIVATE registries
+set -eu
+
+MODELKIT_REF="${MODELKIT_REF:?MODELKIT_REF is required}"
+UNPACK_PATH="${UNPACK_PATH:-/models}"
+MODEL_SUBDIR="${MODEL_SUBDIR:-qwen3}"
+DEST="${UNPACK_PATH}/${MODEL_SUBDIR}"
```
mesutoezdil, line 201: i reproduced this: with MODEL_SUBDIR set to `../something`, DEST resolves outside the mounted volume and the `rm -rf "${DEST}"` further down deletes it. harmless while the value is a literal in the pod spec, but the lab invites templating it. worth a guard on the value?

tutorials/labs/hami-kitops.md
rudrakshkarpe: Correct. Fixed in 36a67c42 by adding `kit login jozu.ml` to the runnable pack and push block in both languages, so the sequence is complete in a fresh shell.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Good catch. Fixed in 36a67c42. The vLLM container now requests 2 CPU and 8 GiB while retaining 8 CPU and 32 GiB limits. Its HAMi extended-resource requests explicitly match the limits, allowing the two workload Pods to fit the documented verification node resources.

tutorials/labs/hami-kitops.md
rudrakshkarpe: You are right. A complete staged tree did not make the `rm` plus `mv` replacement atomic for existing readers. Fixed in 36a67c42 by publishing immutable reference-specific directories and atomically switching the `DEST` symlink with `mv -T`. Previous release directories remain available to processes already using them.

tutorials/labs/hami-kitops.md
rudrakshkarpe: The verified ModelKit contains one config file, but the reusable unpacker should not rely on filesystem order. Fixed in 36a67c42: it now requires exactly one `config.json` and exits with the discovered count when the layout is missing or ambiguous. I also tested the two-config failure path.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Agreed, and thanks for reproducing it. Fixed in 36a67c42 before any derived path or removal is constructed. `UNPACK_PATH` must be exactly the mounted `/models` root, and `MODEL_SUBDIR` must be one safe component with no slash, traversal value, or unsupported character. I tested that `../escape` is rejected before any filesystem mutation.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Correct. Fixed in 36a67c42. The text now states 60000 MiB, or about 58.6 GiB, and the Chinese version uses the same precise conversion.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Agreed. `emptyDir.sizeLimit` constrains use but does not reserve node storage during scheduling. Fixed in 36a67c42 by adding a 20 GiB ephemeral-storage request and limit to each ModelKit init container in both languages.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Fixed. I replaced the stale PR body with the current Lab 16 scope, completed validation, English and Chinese status, reference ModelKit rationale, and the required merge order after #654.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Agreed. Fixed in 36a67c42. Both Dockerfile snippets now use BuildKit `TARGETARCH`, map `amd64` to `x86_64` and `arm64` to `arm64`, and reject unsupported architectures. I built and ran the checksum-verified KitOps v1.11.0 layer successfully for both linux/amd64 and linux/arm64.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Yes. Fixed in 36a67c42 in both languages. The workstation CLI link now points directly to the current `kitops-ml/kitops` repository.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Agreed. An EXIT trap cannot clean up after SIGKILL or node loss. Fixed in 36a67c42 with lock-owner identity, a 15-second heartbeat, conservative stale recovery after 10 minutes, and repeated lock acquisition by waiters. I tested recovery from a deliberately stale lock in a container.

tutorials/labs/hami-kitops.md
rudrakshkarpe: Yes, this needed qualification. Fixed in 36a67c42. The optional shared-PVC section now states that reliable cross-node sharing requires a ReadWriteMany-capable StorageClass and that kind's default local-path provisioner is node-local and unsuitable for the general cross-node case.

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,805 @@
...
+
+# This lock only coordinates Pods when they mount the same shared PVC. With
+# the emptyDir used in this lab, every Pod has an isolated volume and lock.
+# mkdir is atomic, so Pods sharing a PVC do not race to write the same files.
+if ! mkdir "${LOCK}" 2>/dev/null; then
+  echo "[kitunpacker] another unpack in progress, waiting for it to finish..."
+  i=0
+  while [ "${i}" -lt 360 ]; do
+    ready && { echo "[kitunpacker] model became ready"; exit 0; }
+    i=$((i + 1)); sleep 5
```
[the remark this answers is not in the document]
rudrakshkarpe: Fixed in 36a67c42 in both language copies. Waiters now retry atomic lock acquisition, the owner refreshes a heartbeat every 15 seconds, and a lock with no heartbeat progress for 10 minutes is recovered through an atomic rename. Owner identity prevents an old process from deleting a replacement lock.

tutorials/labs/hami-kitops.md
```diff
@@ -0,0 +1,805 @@
...
+#   REGISTRY_URL/USERNAME/PASSWORD  optional creds for PRIVATE registries
+set -eu
+
+MODELKIT_REF="${MODELKIT_REF:?MODELKIT_REF is required}"
+UNPACK_PATH="${UNPACK_PATH:-/models}"
+MODEL_SUBDIR="${MODEL_SUBDIR:-qwen3}"
+DEST="${UNPACK_PATH}/${MODEL_SUBDIR}"
+RAW="${UNPACK_PATH}/.raw-${MODEL_SUBDIR}"
+STAGE="${UNPACK_PATH}/.stage-${MODEL_SUBDIR}-$$"
+LOCK="${UNPACK_PATH}/.lock-${MODEL_SUBDIR}"
```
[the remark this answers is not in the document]
rudrakshkarpe: Fixed in 36a67c42 in both language copies. Validation now runs before constructing `DEST`, `RAW`, `STAGE`, or `LOCK`: `UNPACK_PATH` must be the mounted `/models` root and `MODEL_SUBDIR` must be one safe component. A traversal value such as `../escape` is rejected before filesystem mutation.

rudrakshkarpe:
Maintainer review update in 36a67c42:

- added verified amd64 and arm64 KitOps release selection using `TARGETARCH`
- replaced the delete and rename publication flow with immutable release directories and an atomic destination symlink switch
- require exactly one `config.json`
- added lock-owner identity, heartbeat, retry, and stale-lock recovery
- added strict path validation before any derived path or removal
- updated the KitOps repository link and refreshed the complete PR description
- added Jozu Hub login to the runnable push flow
- added explicit vLLM CPU and memory requests
- reserved 20 GiB of ephemeral storage for each unpack init container
- documented ReadWriteMany requirements for cross-node PVC sharing
- corrected 60000 MiB to about 58.6 GiB
- kept English and Chinese instructions synchronized

Validation completed locally for Markdown lint, Prettier, 73 tests, shell syntax, traversal rejection, ambiguous config rejection, stale-lock recovery, atomic symlink publication, release checksums, and both amd64 and arm64 image builds. The signed-off commit is pushed and DCO passes. All 12 new maintainer threads and both related CodeRabbit threads have replies and are resolved.

The only remaining discussion is the existing question about keeping the verified digest-pinned user ModelKit or moving the same artifact to a HAMi-owned Jozu namespace. The PR body now documents both options.

@mesutoezdil, please review for LGTM after #654 and confirm the preferred ModelKit ownership option.

tutorials/labs/hami-kitops.md
rudrakshkarpe: For the merge-ready version, I kept the concrete user-published ModelKit because it is verified, runnable, and immutable by digest. The PR body now explains that choice and keeps `<your-org>` placeholders in the customization flow. It also records the stronger long-term option: if a HAMi-owned Jozu namespace is provided, I can mirror the same artifact and update the pinned digest without changing the lab workflow.

mesutoezdil:
on the modelkit ownership question: please move the artifact to a HAMi owned jozu namespace and point MODELKIT_REF there, keeping the digest pin.

the reason is that this lab is meant to stay useful for a long time, and right now it depends on a personal account staying in place. same digest, same content, just an owner the project controls.

i went through all twelve threads against 36a67c42 rather than the summary. the two hard ones hold up: `mv -Tf` on the symlink is an atomic rename so there is no window where the destination is missing, and the MODEL_SUBDIR guard rejects the `../` case i reproduced earlier. the arch selection also fails loudly on an unknown TARGETARCH instead of pulling the wrong binary. nice work.

holding lgtm until the namespace move lands. #654 goes first as you suggested.

cc @rootsongjc @archlitchi

rudrakshkarpe:
That makes sense. This is a verified Qwen ModelKit published by a Jozu user, so I kept the exact digest pin. I agree that a HAMi-owned reference is better for this lab. I do not have access to a HAMi-owned Jozu namespace. Could a maintainer provide the target namespace or mirror this artifact and share the new digest-pinned reference? I will update MODELKIT_REF in both pages as soon as it is available.

rootsongjc:
@rudrakshkarpe can you solve the sidebar config conflicts?

rudrakshkarpe:
Resolved. I merged the latest master and kept both Lab 15 and Lab 16 in the sidebar in the correct order. Formatting, lint, tests, and the full site build pass. Thanks! 

@rootsongjc  please check if it's good for lgtm!

mesutoezdil:
@saiyam1814 remove the claude from histroy pls

mesutoezdil:
could you renumber this to lab 17, since #793 is already on lab 16 with an approval and an examples dir while this one has no examples dir to rename?

rudrakshkarpe:
Done. I synced the branch with current `master`, kept #793 as Lab 16, and moved this KitOps tutorial to Lab 17 in both English and Chinese. The sidebar and PR title are updated too. Formatting, lint, all 93 tests, and the full English and Chinese build pass.

mesutoezdil (approved):
added the missing overview entries in both locales myself.

PR #661, opened 2026-07-28
fix: Format blog and case study dates in UTC

### LINKED ISSUE #659: Blog post and case study dates are one day early west of UTC
### Description

Visitors browsing from the Americas see the wrong publish date on every blog post and on `/case-studies`. A post dated 2026-07-16 displays as 7/15/2026. Readers at UTC or east of it see the correct date, which is probably why this hasn't come up.

### Cause

Docusaurus stores front matter `date: "2026-07-16"` as `2026-07-16T00:00:00.000Z`, i.e. UTC midnight. `src/theme/BlogPostItem/Header/index.js` and `src/pages/case-studies.js` each define a local `formatDate` that renders that instant in the browser's timezone:

```js
function formatDate(date, locale) {
  const language = locale === "zh" ? "zh-CN" : "en-US";
  return new Date(date).toLocaleDateString(language, {
    year: "numeric",
    month: "numeric",
    day: "numeric",
  });
}
```

At any negative UTC offset, that lands on the previous calendar day:

```console
$ TZ=America/New_York node -e 'console.log(new Date("2026-07-16T00:00:00.000Z").toLocaleDateString("en-US"))'
7/15/2026
```

### Reproduce

Set your system timezone to America/New_York and open both pages:

- `/blog` lists the post as **July 16, 2026**
- `/blog/vllm-meetup-shanghai-2026-recap` shows **7/15/2026**

Same post, two dates. The list renders through the upstream `BlogPostItem/Header/Info`, which passes `timeZone: 'UTC'`. Our swizzle only replaces the date when `isBlogPostPage` is true, and the hand-rolled `formatDate` it uses there omits the option. The post page's JSON-LD still emits `datePublished: 2026-07-16T00:00:00.000Z`, so the structured data disagrees with the visible date as well.

This also causes a React hydration mismatch, since the date is baked into the static HTML in the build timezone (UTC on Netlify) and then recomputed in the visitor's timezone.

Happy to open a PR to fix this.

Anshumancanrock:

`metadata.date` and `publishedAt` are both UTC midnight, but the `formatDate` helpers in `src/theme/BlogPostItem/Header/index.js` and `src/pages/case-studies.js` rendered them in the viewer's timezone. Anyone at a negative offset got the previous day, so a post dated 2026-07-16 read as 7/15/2026 across the Americas. Added `timeZone: "UTC"` to both.

`formatDate`/`formatDateRange` in `src/utils/date.js` have the same omission, so I fixed those too. Nothing imports them today, but they're exported and would reintroduce this. The other helpers in that file take Dates the events calendar already built in local time, or render a wall-clock time, so those deliberately stay local.

Nothing changes for readers at UTC or east of it. I diffed every rendered date in `build/` before and after and they're identical. Ran the fixed formatters under UTC, Shanghai, Kolkata, Berlin, New York, LA and Honolulu and they now all agree, which also clears the hydration mismatch since the static output can no longer differ from what the client computes.

I didn't move this to `useDateTimeFormat`, which is what upstream and `BlogArchivePage` use. It doesn't default to UTC either, so it'd be this same change plus a refactor, and it's a hook so `formatDate` would have to move into the component. Happy to do that separately if you'd rather have the consistency.

Fixes #659

**Checklist**:

- [ ] Chinese translation updated if English docs changed (or noted why not) — no doc content changed, formatting only
- [ ] Commits are signed off (`git commit -s`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected date displays to consistently use UTC, preventing dates from shifting based on the viewer’s local time zone.
  * Standardized date and date-range formatting across case studies, blog posts, and event calendars.

### DIFF
--- a/src/pages/case-studies.js
+++ b/src/pages/case-studies.js
@@ -10,6 +10,7 @@ function formatDate(date, locale) {
     year: "numeric",
     month: "short",
     day: "numeric",
+    timeZone: "UTC",
   });
 }
 
--- a/src/theme/BlogPostItem/Header/index.js
+++ b/src/theme/BlogPostItem/Header/index.js
@@ -15,6 +15,7 @@ function formatDate(date, locale) {
     year: "numeric",
     month: "numeric",
     day: "numeric",
+    timeZone: "UTC",
   });
 }
 
--- a/src/utils/date.js
+++ b/src/utils/date.js
@@ -5,6 +5,7 @@ export function formatDate(dateStr, locale) {
     year: "numeric",
     month: "long",
     day: "numeric",
+    timeZone: "UTC",
   }).format(new Date(dateStr));
 }
 
@@ -13,9 +14,11 @@ export function formatDateRange(startStr, endStr, locale) {
     year: "numeric",
     month: "long",
     day: "numeric",
+    timeZone: "UTC",
   }).formatRange(new Date(startStr), new Date(endStr));
 }
 
+// These take local Dates from the events calendar, so no UTC here.
 export function formatDay(date, locale) {
   return date.toLocaleDateString(LOCALE_MAP[locale], { day: "numeric" });
 }


Anshumancanrock:
Hi @rootsongjc, could you please take a look at this PR when you get a chance? Thanks!

rootsongjc (approved):
/lgtm

PR #679, opened 2026-07-29
ci: add issue templates for website bugs, enhancements and good first issues

mesutoezdil:
## Why

Two things were wrong with issue labeling here.

**The existing automation never worked.** Both current templates declare `labels: kind/documentation`, but that label did not exist in this repository. GitHub drops an unknown label silently, with no warning on the issue or in the template. Since `blank_issues_enabled: false` forces everyone through a template, every issue filed here should have arrived labeled, and none did. That is most of why 18 of 21 open issues had no label before the recent manual pass.

**Three issue types had no template.** The repository only offered "Documentation issue" and "Documentation improvement", both about docs content. Site bugs (#659 wrong dates, #664 broken code blocks), enhancements (#658 SEO, #662 image sizing, #666 i18n) and good first issues had nowhere to go.

## What changed

Labels created first, matching HAMi's names, colors and descriptions:

| Label | Color |
|---|---|
| `kind/documentation` | `0075ca` |
| `kind/bug` | `d73a4a` |
| `kind/feature` | `C5DEF5` |

Labels live in repository settings rather than in the tree, so they are not part of this diff.

Templates added:

| Template | Label applied |
|---|---|
| `good-first.md` | `good first issue` |
| `site-bug.md` | `kind/bug` |
| `enhancement.md` | `kind/feature` |

Every template in the directory now maps to a label that exists:

```
doc-improvement.md -> kind/documentation
doc-issue.md       -> kind/documentation
enhancement.md     -> kind/feature
good-first.md      -> good first issue
site-bug.md        -> kind/bug
```

`good-first.md` mirrors the one in HAMi, including the `/assign` instruction and the note that good first issues are reserved for first-time contributors.

## Effect

New issues arrive labeled without anyone doing a manual pass. The `good first issue` filter, which is how newcomers find an entry point, fills itself as maintainers file tasks under that template.

`npm run format:check` passes.

## Summary by CodeRabbit

* **Documentation**
  * Added structured issue templates for enhancement requests, good first issues, and website bug reports.
  * Templates provide guided prompts and relevant metadata to improve the consistency and completeness of submitted issues.

PR #681, opened 2026-07-29
ci: label pull requests from their changed paths

mesutoezdil:
## Why

Issue labels were applied in a single manual pass. Pull requests carry no `area/*` label at all, and nothing keeps either up to date as new work arrives.

Mapping a changed path to an area needs no judgement, so it can run on its own. Priority and difficulty still cannot, and are left alone.

## What it does

`.github/labeler.yml` maps paths to the `area/*` labels.

| Label | Paths |
|---|---|
| `area/docs` | `docs/`, `versioned_docs/`, `versioned_sidebars/`, `tutorials/`, `sidebars*.js`, `versions.json` |
| `area/ui` | `src/` |
| `area/blog` | `blog/` |
| `area/i18n` | `i18n/` |
| `area/edge-functions` | `netlify/`, `netlify.toml` |
| `area/ci-infra` | `.github/`, `scripts/`, `package*.json`, lint and format config |
| `area/seo` | `static/robots.txt`, `static/llms.txt`, `docusaurus.config.js` |

`area/ui` is new, created to cover `src/` (React components, pages, theme, css), which is the largest source directory here and had no area label. It uses the same colour as the rest of the `area/*` family. Every other label already existed.

## Dry run against merged pull requests

Ran the rules against the actual changed file lists:

| PR | Files | Result |
|---|---|---|
| #679 | `.github/ISSUE_TEMPLATE/*` | `area/ci-infra` |
| #678 | `src/pages/index.js` | `area/ui` |
| #677 | `.github/workflows/docs-health.yml` | `area/ci-infra` |
| #673 | `package.json`, `package-lock.json` | `area/ci-infra` |
| #655 | `i18n/…`, `tutorials/…`, `sidebars-tutorials.js` | `area/i18n`, `area/docs` |

Every merged pull request checked now lands on at least one area.

## Notes

`pull_request_target` is used because a `pull_request` run from a fork gets a read-only token and cannot add labels. The workflow never checks out or executes pull request code, it only reads the changed file list, so the elevated token is not exposed to contributor code. The action is pinned by commit SHA, matching the convention in `docs-health.yml`.

`sync-labels: false` means labels are only added, never removed, so a manually applied label is not stripped on the next push.

`npm run format:check` passes.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,46 @@
+area/docs:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "docs/**"
+          - "versioned_docs/**"
+          - "versioned_sidebars/**"
+          - "tutorials/**"
+          - "sidebars.js"
+          - "sidebars-tutorials.js"
+          - "versions.json"
+
+area/ui:
+  - changed-files:
+      - any-glob-to-any-file: "src/**"
+
+area/blog:
+  - changed-files:
+      - any-glob-to-any-file: "blog/**"
+
+area/i18n:
+  - changed-files:
+      - any-glob-to-any-file: "i18n/**"
+
+area/edge-functions:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "netlify/**"
+          - "netlify.toml"
+
+area/ci-infra:
+  - changed-files:
+      - any-glob-to-any-file:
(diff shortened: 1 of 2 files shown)


rootsongjc (approved):
/lgtm

PR #684, opened 2026-07-29
docs: audit supported devices matrix and clarify support status

Adii-45:
## What type of PR is this?

Documentation enhancement

---

## What this PR does / why we need it

This PR addresses the first documentation slice for issue #672 by auditing the supported devices matrix.

Changes include:

- Added missing AWS Neuron entry
- Added missing Biren entry in the current (Next) docs
- Updated Teco status
- Added a Status column and support-status definitions
- Synced Chinese translation
- Added missing Chinese sidebar translation key for Biren

---

## Which issue(s) this PR fixes

Part of #672

---

## Checklist

- [x] npm run lint and npm run format:check pass
- [x] npm run build succeeds for both en and zh
- [x] Chinese translation updated
- [x] Commits are signed off (`git commit -s`)

---

## AI Assistance

I consulted Claude Code while working on this documentation update (primarily for repository navigation, understanding the existing documentation structure, and reviewing wording). All changes, verification of the documentation, and the final content were reviewed and authored by me!

## Summary by CodeRabbit

* **Documentation**
  * Updated the supported-device table with support status labels and capability details.
  * Added support information for AWS Neuron and Biren devices.
  * Clarified DPU support and isolation capabilities, including its validation status.
  * Added definitions for support statuses and capability columns.
  * Added device setup guide links.
  * Updated Chinese documentation and added the translated device-management sidebar category.

PR #682, opened 2026-07-29
perf(media): update image handling and dependencies

vivek-gite:
**What type of PR is this?**

feature

**What this PR does / why we need it**:

Implements the responsive-image work from #662 so visitors download image candidates appropriate to the rendered size instead of always receiving large source files.

- Adds a shared `ResponsiveImage` component that renders `picture`, `srcSet`, `sizes`, intrinsic dimensions, lazy loading, and async decoding.
- Generates responsive WebP and original-format variants for the SZZT adopter logo and KubeCon booth banner.
- Updates the logo wall and relevant image rendering paths while preserving lightbox behavior, alt text, and figure numbering.
- Adds build-time image-size validation. The full repository scan runs in CI rather than delaying `npm start`.

This reduces image transfer and decode work on mobile while reserving image space to avoid layout shift.

**Which issue(s) this PR fixes**:

Fixes #662

**AI assistance**:

I used AI assistance to help me explore the codebase and review code. But solution was fully authored, manually reviewed, tested by me and take responsibility for all changes in this PR.

**Checklist**:

- [x] `npm run lint` and `npm run format:check` pass
- [x] `npm run build` succeeds for both `en` and `zh`
- [x] Chinese translation updated if English docs changed (or noted why not)
- [x] Commits are signed off (`git commit -s`)

## Summary by CodeRabbit

* **New Features**
  * Added responsive image delivery with optimized WebP and size-specific variants.
  * Updated supported blog and logo images to adapt to screen size and improve loading performance.
  * Preserved access to full-resolution images through the lightbox viewer.
* **Bug Fixes**
  * Improved figure captions for images rendered with responsive picture elements.
  * Corrected image sizing within support sections.
* **Chores**
  * Added automated image-size validation during documentation builds to help prevent oversized assets.

PR #687, opened 2026-07-30
chore: pin CodeRabbit label vocabulary to the kind/* scheme

mesutoezdil:
**What type of PR is this?**

/kind cleanup

Follow-up to the label discussion in #679. The flat `bug`, `enhancement` and `documentation` labels were deleted and the four issues that carried `enhancement` are on `kind/feature`, but the delete does not hold on its own: CodeRabbit auto-applies labels here and, with no config in the repo, infers them from past PRs. Adding a missing label through the GitHub API recreates it, so `enhancement` came straight back (see the label timeline on this PR, #682 and #684).

This pins the vocabulary:

- suggestions restricted to `kind/bug`, `kind/feature`, `kind/documentation` and `kind/cleanup`
- the four marked mutually exclusive, so a PR cannot end up with two type labels
- `auto_apply_labels: true` keeps the current behavior, only with the right labels

`area/*` is left to the path globs in `.github/labeler.yml` and `priority/*` stays a maintainer call, so this file does not touch either.

**On `kind/cleanup`**: `.github/PULL_REQUEST_TEMPLATE.md` offers `/kind bug`, `/kind cleanup`, `/kind documentation` and `/kind feature`, but `kind/cleanup` did not exist as a label. Prow silently rejected it, which is the same class of bug #679 describes for the issue templates. The label now exists, so it belongs in this list too.

**Note on scope**: CodeRabbit only reads `.coderabbit.yaml` from the base branch for open source repos, so this has no effect until it lands on `master`. Until then `enhancement` keeps reappearing on new PRs. Deleting that label again after merge is the last step.

Relates to #679

**Checklist**:

Notes on the checklist:

- `prettier --check` passes on the new file, and the config validates against the CodeRabbit v2 config schema (`schema.v2.json`).
- `Docs Health Check / Lint, Format, Build & Links` passed on CI, which covers lint, format and the build for both locales. `.coderabbit.yaml` is read by CodeRabbit only, it is not an input to Docusaurus.
- No English docs changed, so no translation work.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,27 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+#
+# Label vocabulary for CodeRabbit. Without this file CodeRabbit infers labels
+# from past pull requests, which is how the flat `enhancement` label kept coming
+# back after it was deleted. With `labeling_instructions` set, CodeRabbit only
+# suggests labels from the list below.
+#
+# Scope is limited to `kind/*` on purpose. `area/*` is applied deterministically
+# by path globs in .github/labeler.yml, and `priority/*` is a maintainer call.
+
+reviews:
+  labeling_instructions:
+    - label: kind/bug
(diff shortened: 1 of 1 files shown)


rootsongjc (approved):
/lgtm

PR #688, opened 2026-07-30
test: add unit tests for the markdown edge function

Anshumancanrock:
**What type of PR is this?**

/kind cleanup

`netlify/edge-functions/markdown-negotiation.js` converts every page to Markdown for clients sending `Accept: text/markdown`. It is around 290 lines of regex, it runs on every request, and it had no tests. It already regressed once in #665, where code blocks and tables came out mangled.

This PR adds 69 tests. They drive the exported handler rather than the internals, so the edge function itself is unchanged. Coverage is Accept parsing, the asset-extension skip, response headers, the `config` route, the document preamble, lists, tables, code fences, entities and whitespace. The two behaviours #665 had to fix are pinned specifically, since neither is obvious from reading the code: code blocks and tables are parked before whitespace normalization so significant indentation survives, and fences widen past any backtick run in the body.

It uses the built-in `node:test` runner, so there are no new dependencies and the lockfile does not move. Adding Jest or Vitest would pull a few hundred packages into a docs site.

To check the tests actually catch things, I injected 46 regressions into the edge function one at a time and confirmed every one turns the suite red. I also ran the function over all 1027 pages of a production build: no crashes, no HTML leaking into the output, no unbalanced fences.

#### engines bump

`engines.node` moves from `>=20.18.1` to `>=20.19.0`. The edge function is ESM in a `.js` file while the package is CommonJS, so importing it needs module-syntax detection, which 20.18.1 does not have. On 20.18.1 the import fails with `SyntaxError: Unexpected token 'export'`; 20.19.0 and 24.x both pass. CI already uses `node-version: 20` and `netlify.toml` pins `NODE_VERSION = "20"`, both of which resolve to the latest 20.x, and `.npmrc` does not set `engine-strict`, so nothing else needs to change.

Tests live in `test/` rather than beside the source because every top-level file under `netlify/edge-functions/` is deployed as its own edge function.

#### Two bugs turned up while writing these. Not fixed here, happy to take both as follow-ups:

1. Nested lists get flattened. The `<li>` pattern is non-greedy, so the first `</li>` closes the nested item instead of the outer one. Sub-items are promoted to top-level steps and everything after them is renumbered. This hits all 21 built pages that nest a list inside an `<ol>`, across 5 guides: `gpu-virtualization`, `enable-cambricon-mlu-sharing`, `enable-metax-gpu-schedule`, `grafana-dashboard` and `hami-webui-development-guide`. On the WebUI dev guide, five sub-bullets come out as steps 2-5 and the real step 2 is served as step 6.
2. Attribute values containing `>` leak. Tag scanning uses `[^>]*`, so it stops at the first `>` even inside a quoted attribute, and `<p title="a > b">Hello</p>` renders as `b">Hello`.

**Checklist**:

- [x] Chinese translation updated if English docs changed (no docs content changed, so nothing to translate)

**AI assistance**:

Drafted with AI assistance (Claude) and planned, reviewed, and verified by me.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "write-translations": "docusaurus write-translations",
     "write-heading-ids": "docusaurus write-heading-ids",
     "fetch-changelog": "node src/scripts/fetch-changelog.js",
-    "test": "node --test",
+    "test": "node --test test/*.test.mjs",
     "lint:md": "markdownlint 'docs/**/*.md' 'tutorials/**/*.md' 'blog/**/*.md' 'i18n/zh/docusaurus-plugin-content-docs-tutorials/current/**/*.md' 'i18n/zh/docusaurus-plugin-content-docs/current/**/*.md' 'i18n/zh/docusaurus-plugin-content-blog/**/*.md' --ignore CHANGELOG.md --ignore changelog --ignore versioned_docs --ignore node_modules",
     "lint": "npm run lint:md",
     "format:check": "prettier --check .",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "version": "0.0.1",
   "engines": {
-    "node": ">=20.18.1"
+    "node": ">=20.19.0"
   },
   "private": true,
   "scripts": {
@@ -24,12 +24,13 @@
     "write-translations": "docusaurus write-translations",
     "write-heading-ids": "docusaurus write-heading-ids",
     "fetch-changelog": "node src/scripts/fetch-changelog.js",
+    "test": "node --test",
     "lint:md": "markdownlint 'docs/**/*.md' 'tutorials/**/*.md' 'blog/**/*.md' 'i18n/zh/docusaurus-plugin-content-docs-tutorials/current/**/*.md' 'i18n/zh/docusaurus-plugin-content-docs/current/**/*.md' 'i18n/zh/docusaurus-plugin-content-blog/**/*.md' --ignore CHANGELOG.md --ignore changelog --ignore versioned_docs --ignore node_modules",
     "lint": "npm run lint:md",
     "format:check": "prettier --check .",
     "format": "prettier --write .",
     "check-links": "DOCUSAURUS_SITE_URL=http://127.0.0.1:3000 npm run build && bash scripts/check-links.sh",
-    "check:all": "npm run lint && npm run format:check && npm run check-links"
+    "check:all": "npm run lint && npm run format:check && npm test && npm run check-links"
   },
   "dependencies": {
     "@docusaurus/core": "^3.10.1",
--- a/test/markdown-negotiation.test.mjs
+++ b/test/markdown-negotiation.test.mjs
@@ -216,8 +216,8 @@ describe("document preamble", () => {
 
   // Docusaurus always emits <main>; this is the fallback for anything that does
   // not. The whole document is converted, so text from <head> ends up in the
-  // body — acceptable for a path that production never takes, but pinned so the
-  // fallback is not mistaken for clean extraction.
+  // body. That is acceptable for a path production never takes, but it is
+  // pinned so the fallback is not mistaken for clean extraction.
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -24,12 +24,13 @@
     "write-translations": "docusaurus write-translations",
     "write-heading-ids": "docusaurus write-heading-ids",
     "fetch-changelog": "node src/scripts/fetch-changelog.js",
+    "test": "node --test",
```
mesutoezdil: `node --test` with no path walks the whole repo, i tried it and it picks up test files under `build/` too, pls scope it to `node --test test/`.

test/markdown-negotiation.test.mjs
```diff
@@ -0,0 +1,558 @@
...
+
+  it("keeps a body H1 that is genuinely different", async () => {
+    const markdown = await render(page("<h1>Prerequisites</h1>", { title: "Installation" }));
+    assert.match(markdown, /^# Installation$/m);
+    assert.match(markdown, /^# Prerequisites$/m);
+  });
+
+  // Docusaurus always emits <main>; this is the fallback for anything that does
+  // not. The whole document is converted, so text from <head> ends up in the
+  // body — acceptable for a path that production never takes, but pinned so the
```
mesutoezdil: em dash isn't allowed in our repo, pls use a plain `-` here (the one on line 511 has to stay, it's asserting the `&mdash;` output).

mesutoezdil (commented):
the pr description doesn't follow the repo template, pls fill the 'what this PR does / why we need it' section. there is also no `/kind` line, so prow can't put a type label on it.

no ai assistance disclosure is present. if any ai tool was used, it must be disclosed per CONTRIBUTING.md: https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md#ai-assistance-notice

package.json
Anshumancanrock: Thanks, agreed that bare `node --test` was too broad.

I tried `node --test test/` first, but it only works on Node 20:

```
Node 20.19.0  ->  69 pass
Node 22.14.0  ->  0 pass, 1 fail
Node 24.15.0  ->  Error: Cannot find module '.../website/test'
```

Node 22 changed positional arguments to glob patterns, so a bare directory is
resolved as a module to execute rather than a directory to scan. CI would stay
green since it pins Node 20, but contributors on newer runtimes would hit this
locally.

I've gone with `cd test && node --test`. It keeps the scoping you asked for,
doesn't depend on shell glob expansion (so it works on Windows as well), and
still discovers anything added under `test/` subdirectories.

Happy to change it if you'd prefer a different approach.

Anshumancanrock:
@mesutoezdil thanks for the review. I have rewrote the description on the template with /kind cleanup and addressed all your comments. Please take a look now.

package.json
mesutoezdil: > Thanks, agreed that bare `node --test` was too broad.
> 
> I tried `node --test test/` first, but it only works on Node 20:
> 
> ```
> Node 20.19.0  ->  69 pass
> Node 22.14.0  ->  0 pass, 1 fail
> Node 24.15.0  ->  Error: Cannot find module '.../website/test'
> ```
> 
> Node 22 changed positional arguments to glob patterns, so a bare directory is resolved as a module to execute rather than a directory to scan. CI would stay green since it pins Node 20, but contributors on newer runtimes would hit this locally.
> 
> I've gone with `cd test && node --test`. It keeps the scoping you asked for, doesn't depend on shell glob expansion (so it works on Windows as well), and still discovers anything added under `test/` subdirectories.
> 
> Happy to change it if you'd prefer a different approach.

Reminder: Answers must be written by human being. You can view the relevant rule here.
https://github.com/Project-HAMi/HAMi/blob/master/CONTRIBUTING.md#contribution-gates
"4. Review replies. The reply you post must be written by you and must address the specific point raised. Verbatim or canned AI replies, or replies that do not engage the comment, lead to the PR being closed."

PR #696, opened 2026-08-01
fix(changelog): normalize CRLF before parsing so committer regex matches

### LINKED ISSUE #695: Changelog plugin drops contributor attribution and crashes the zh build on Windows (CRLF checkout)
**Page URL**:
- Not a page rendering bug - this is in the build tooling: `src/plugins/changelog/index.js`, hit via `npm run build` (the command CONTRIBUTING.md tells contributors to run locally before pushing).

**What happened**:

On a Windows checkout with the default `core.autocrlf=true`, `npm run build` crashes while building the `zh` locale:

```
[ERROR] Error: Unable to build website for locale zh.
    ...
    [cause]: Error: Processing of blog source file path=v2.7.0.md failed.
    [cause]: Error: Can't reference blog post authors by a key (such as 'abstractmj') because no authors map file could be loaded.
    Please double-check your blog plugin config (in particular 'authorsMapPath'), ensure the file exists at the configured path, is not empty, and is valid!
```

The `en` build finishes fine right before it, so it's easy to miss anything's wrong until the `zh` pass blows up.

Root cause: `src/plugins/changelog/index.js` builds `changelog/source/authors.json` by regex-parsing the `#### Committers:` section of `CHANGELOG.md`:

```js
let authors = content.match(/#### Committers:[\s\S]*$/);
if (authors) {
  authors = authors[0]
    .match(/- .*?\n/g)   // <-- line 66
    ...
```

`.` in a JS regex doesn't match `\r`. Once `CHANGELOG.md` is checked out with CRLF line endings, every committer line looks like `- Kevinz857 (...)\r\n`, and `/- .*?\n/g` never matches (it needs a literal `\n` right after the lazy `.*?`, but hits `\r` first). The match returns `null`, `authorsMap` stays `{}`, and `authors.json` gets written out empty for every release, silently, no error at that point.

That alone just drops contributor bylines from the generated (English) changelog pages. It turns into a hard build failure because `i18n/zh/docusaurus-plugin-content-blog-changelog/v2.7.0.md` (and the other localized changelog entries) are tracked files that still have an explicit `authors:` frontmatter list from when they were translated. With `authorsMap` empty, the blog plugin can't resolve those aliases and throws, taking down the whole `zh` locale build.

`core.autocrlf=true` is Git's own recommended default for Windows, so this isn't an exotic local misconfiguration - it's what a new Windows contributor gets out of the box from a standard Git-for-Windows install. There's no `.gitattributes` entry forcing `CHANGELOG.md` to stay LF on checkout. CI runs on Linux so it never sees this, which is presumably why it's gone unnoticed.

Repro:

```
git config --global core.autocrlf true   # default on Windows
git clone https://github.com/Project-HAMi/website && cd website
npm ci
npm run build
```

Watch it succeed for `en` and then die on `zh` with the error above. Or, without touching git config: `cat changelog/source/authors.json` after any build - it's `{}` regardless of locale.

**What you expected**:

`npm run build` to succeed the same way on Windows as it does in CI, and for changelog contributor attribution to survive regardless of the checked-out line-ending style.

Possible fix:
- Make the committer-line regex CRLF-tolerant (`/-.*?\r?\n/g`, or normalize `content.replace(/\r\n/g, '\n')` before parsing) in `src/plugins/changelog/index.js`.
- Add a `.gitattributes` rule so `CHANGELOG.md` always checks out with LF regardless of the contributor's `core.autocrlf` setting.

**Browser and locale** (if it is a rendering issue):
- N/A - build-time bug, not a rendering issue. Affects the `zh` locale build output specifically.

**Acceptance criteria**

- [ ] `npm run build` succeeds for both locales on a fresh Windows checkout with `core.autocrlf=true`.
- [ ] `changelog/source/authors.json` is populated with real contributor entries after a build, not `{}`.
- [ ] `.gitattributes` pins `CHANGELOG.md` to LF line endings.

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

`npm run build` crashes on the `zh` locale on a Windows checkout with the default `core.autocrlf=true`. The changelog plugin extracts committer names from `CHANGELOG.md` with a regex that needs a literal `\n` right after each line; on a CRLF checkout every line ends `\r\n`, so the regex never matches, `authorsMap` stays empty, and `authors.json` gets written out blank. That silently drops contributor bylines from the generated English changelog pages, and it takes down the `zh` build entirely, because the tracked `i18n/zh/docusaurus-plugin-content-blog-changelog/*.md` files still carry an explicit `authors:` list from when they were translated, and the blog plugin can't resolve those aliases against an empty map.

This normalizes `CHANGELOG.md`'s line endings right after reading it, so parsing behaves the same no matter how the file was checked out, and pins `CHANGELOG.md` to LF in `.gitattributes` so a different local git config can't reintroduce this.

Verified locally: before the fix, `changelog/source/authors.json` was `{}` after any build and `npm run build` died mid-way through the `zh` pass with `Can't reference blog post authors by a key (such as 'abstractmj')...`. After the fix, `authors.json` has 84 real contributor entries and both `en` and `zh` builds complete cleanly.

Fixes #695

**Checklist**:

- [x] Chinese translation updated if English docs changed (or noted why not) - N/A, no doc content changed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved changelog processing by consistently handling Windows-style line endings.
  * Changelog sections now parse reliably across supported operating systems.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,3 +6,5 @@
 *.mp4 filter=lfs diff=lfs merge=lfs -text
 *.pdf filter=lfs diff=lfs merge=lfs -text
 *.webp filter=lfs diff=lfs merge=lfs -text
+
+CHANGELOG.md text eol=lf
--- a/src/plugins/changelog/index.js
+++ b/src/plugins/changelog/index.js
@@ -138,7 +138,7 @@ export default async function ChangelogPlugin(context, options) {
     ...blogPlugin,
     name: 'changelog-plugin',
     async loadContent() {
-      const fileContent = await fs.readFile(changelogPath, 'utf-8');
+      const fileContent = (await fs.readFile(changelogPath, 'utf-8')).replace(/\r\n/g, '\n');
       const sections = fileContent
         .split(/(?=\n## )/)
         .map(processSection)


bhuvan-somisetty:
Hey @rootsongjc @mesutoezdil, whenever you have a minute could you take a look at this one? Small, focused fix — root-caused and verified locally (before/after on `authors.json` and both locale builds), details in the PR description. Thanks!

bhuvan-somisetty:
Thanks for the review! Glad the two-part fix (gitattributes + runtime CRLF normalization) checks out. Nothing else to change on my end let me know if you'd like anything tweaked.

rootsongjc (approved):
/lgtm

bhuvan-somisetty:
Thanks @rootsongjc!

PR #700, opened 2026-08-02
fix: scope changelog plugin state per loadContent and derive figure locale from pathname (fixes #699)

### LINKED ISSUE #699: [Bug/Architecture] Build-time plugin state pollution in changelog generator and client-side figure caption race condition
**Page URL**: 
- Build pipeline: `npm run build` / `changelog/source/`
- Client module: `/docs`, `/zh/docs`, `/docs/*`, `/zh/docs/*`, and `/blog/*`

**What happened**:

While auditing the website build pipeline and client-side runtime modules, we identified two systemic defects affecting site build consistency, internationalization (i18n) accuracy, and figure rendering:

1. **Global state leak in `changelog-plugin`**: Module-level variables in the changelog plugin (`publishTimes` and `authorsMap` in `src/plugins/changelog/index.js`) persist across locale build passes within the same Node process during `npm run build`. On multi-locale builds, `publishTimes.has(...)` evaluates to `true` for every entry during the `zh` pass, causing the `while` loop to decrement `hour` (`20` → `19` → `18`...). This causes release dates, changelog post order, and RSS/Atom feed timestamps for Chinese to drift and desynchronize from English.
2. **Client-side race condition, locale desync, and broken anchor links in `imageFigureNumber.js`**: `onRouteDidUpdate` uses triple `setTimeout` timers (100ms, 500ms, 1000ms) and reads `document.documentElement.lang`. During SPA navigation between `/docs/` and `/zh/docs/`, `lang` is updated by Docusaurus after route transition. Early timeout calls read stale `lang` attributes, outputting English captions on Chinese pages or vice versa. Additionally, root doc paths (`/docs` or `/zh/docs`) were missed by strict `/\/docs\//` regex, and wrapping unlinked images detached `` from parent `<a>` tags.

---

**What you expected**:

1. `changelog-plugin` state (`publishTimes`, `authorsMap`) is scoped to `loadContent()` / plugin instance execution so multi-locale builds produce identical, deterministic timestamps and clean author maps across all locales.
2. `imageFigureNumber.js` determines locale deterministically from location pathname (e.g. `/zh/` prefix), covers root doc routes, preserves anchor wrappers around images, and uses clean idempotency guards to prevent duplicate captions.

---

**Browser and locale**:
- Node.js build environment (Node 20+)
- All browsers (Chrome, Firefox, Safari) across `en` and `zh` locales

---

#### 1. Module-level global state in `changelog-plugin`
- **Location**: [`src/plugins/changelog/index.js`](https://github.com/Project-HAMi/website/blob/master/src/plugins/changelog/index.js#L17-L21)
  ```javascript
  const publishTimes = new Set();
  const authorsMap = {};
  ```
  During production builds (`npm run build`), Docusaurus runs `loadContent()` sequentially for `en` and `zh` within the same process. Because `publishTimes` is not cleared between locale passes, `zh` detects `en` timestamps as collisions and shifts Chinese changelog timestamps backward.

#### 2. Client-side DOM race condition and link detachment in `imageFigureNumber.js`
- **Location**: [`src/client/imageFigureNumber.js`](https://github.com/Project-HAMi/website/blob/master/src/client/imageFigureNumber.js)
  `setTimeout` polling reads mutable `document.documentElement.lang` which lags behind SPA route transitions, missed root routes like `/docs`, and detached images from anchor wrappers.

---

### Acceptance Criteria

- [x] `changelog-plugin` state is scoped per plugin instance / `loadContent()` call.
- [x] `imageFigureNumber.js` derives locale from pathname, covers root doc paths, preserves anchor tags around linked images, and enforces idempotent DOM updates.
- [x] `npm run check:all` passes cleanly.

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

This PR resolves build-time state pollution in the changelog generator plugin and client-side race conditions / anchor detachment in figure caption generation:

1. **Changelog plugin state scoping (`src/plugins/changelog/index.js`)**:
   - `publishTimes` (`Set`) and `authorsMap` (`Object`) were defined as module-level global singletons. During production multi-locale builds (`npm run build`), Docusaurus runs `loadContent()` sequentially for `en` then `zh` in the same Node process. Because `publishTimes` was never reset between locale runs, the `zh` pass detected all timestamps from `en` as existing collisions. The dedup loop (`while (publishTimes.has(...)) hour -= 1`) decremented release hours for every entry, shifting Chinese changelog dates, RSS/Atom feed timestamps, and post sorting backward relative to English.
   - **Fix**: Moved `publishTimes` and `authorsMap` instantiation inside `loadContent()` and passed them down to `processSection`. Each build pass now starts with a clean slate, ensuring deterministic and matching timestamps across all locales.

2. **Client-side figure caption race condition & anchor preservation (`src/client/imageFigureNumber.js`)**:
   - `onRouteDidUpdate` scheduled triple `setTimeout` timers (100ms, 500ms, 1000ms) and read `document.documentElement.lang` to decide between `Figure` and `图` caption prefixes. Because Docusaurus updates `<html lang="...">` asynchronously after route transitions complete, early timeout callbacks read stale `lang` values, outputting English captions on Chinese pages or vice versa. Additionally, root doc paths (`/docs` or `/zh/docs`) were missed by strict regex, and wrapping unlinked images detached `` elements from parent `<a>` link tags.
   - **Fix**: Derived locale directly from `location.pathname` (checking for `/zh/`), expanded route regex to match root doc routes `/(?:^|\/)(?:blog|docs)(?:$|\/)/`, preserved parent `<a>` tags inside `<figure>` wrappers for linked images, replaced triple `setTimeout` calls with a single `requestAnimationFrame` tick, and ensured `<figcaption>` text is always updated dynamically.

Fixes #699

**Checklist**:

- [x] Chinese translation updated if English docs changed (N/A — code bug fix)

## Summary by CodeRabbit

* **New Features**
  * Figure captions now use the page’s route locale, including documentation and blog root paths.
  * Captions are generated after images finish loading, improving coverage for asynchronously loaded images.
  * Clickable images retain their links when placed inside figures.

* **Bug Fixes**
  * Icon images are more reliably excluded from automatic figure captioning.
  * Existing figures and captions now update consistently with the current figure number, locale, and alternative text.
  * Changelog generation now keeps content from separate locale builds isolated.

### DIFF
--- a/src/client/imageFigureNumber.js
+++ b/src/client/imageFigureNumber.js
@@ -16,8 +16,8 @@ function localeFromPathname(pathname) {
 }
 
 export function onRouteDidUpdate({ location }) {
-  // Only run on blog and doc pages
-  if (!location.pathname.match(/\/blog\//) && !location.pathname.match(/\/docs\//)) {
+  // Only run on blog and doc pages (including root doc paths like /docs or /zh/docs)
+  if (!/(?:^|\/)(?:blog|docs)(?:$|\/)/.test(location.pathname)) {
     return;
   }
 
@@ -60,12 +60,15 @@ function addFigureNumbers(locale) {
 
     figureCount++;
 
-    // Wrap in <figure> if not already wrapped
+    // Wrap in <figure> if not already wrapped. If the image is inside an anchor link,
+    // move the anchor element into the figure to preserve clickable links.
     let figure = img.closest("figure");
     if (!figure) {
       figure = document.createElement("figure");
-      img.parentNode.insertBefore(figure, img);
-      figure.appendChild(img);
+      const targetElement =
+        img.parentElement && img.parentElement.tagName === "A" ? img.parentElement : img;
+      targetElement.parentNode.insertBefore(figure, targetElement);
+      figure.appendChild(targetElement);
     }
 
     // Find or create the <figcaption>, always overwriting its text so that
--- a/src/client/imageFigureNumber.js
+++ b/src/client/imageFigureNumber.js
@@ -2,28 +2,48 @@
  * Client module to automatically add figure numbers and captions to images
  * based on their alt text in blog posts and docs.
  */
-export function onRouteDidUpdate({ location, previousLocation }) {
+
+/**
+ * Derive the locale from the current pathname so the caption prefix is always
+ * consistent with the page the user navigated *to*, not with whatever stale
+ * value `document.documentElement.lang` happens to hold during the transition.
+ *
+ * @param {string} pathname
+ * @returns {"zh" | "en"}
+ */
+function localeFromPathname(pathname) {
+  return pathname.startsWith("/zh/") || pathname === "/zh" ? "zh" : "en";
+}
+
+export function onRouteDidUpdate({ location }) {
   // Only run on blog and doc pages
   if (!location.pathname.match(/\/blog\//) && !location.pathname.match(/\/docs\//)) {
     return;
   }
 
-  // Wait for DOM to be ready with multiple attempts
-  setTimeout(() => addFigureNumbers(), 100);
-  setTimeout(() => addFigureNumbers(), 500);
-  setTimeout(() => addFigureNumbers(), 1000);
+  const locale = localeFromPathname(location.pathname);
+
+  // A single rAF tick is sufficient: Docusaurus has already committed the new
+  // page content to the DOM by the time onRouteDidUpdate fires. Using rAF
+  // instead of multiple staggered setTimeouts avoids duplicate caption
+  // injections when routes are navigated quickly.
+  requestAnimationFrame(() => addFigureNumbers(locale));
 }
 
-function addFigureNumbers() {
+/**
+ * @param {"zh" | "en"} locale
+ */
+function addFigureNumbers(locale) {
   // Find all images in markdown content
   const articleContent = document.querySelector("article");
   if (!articleContent) {
     return;
   }
 
-  // Find all images that are not in header or footer
+  const prefix = locale === "zh" ? "图" : "Figure";
+
+  // Find all images that are not logos or small icons
   const images = Array.from(articleContent.querySelectorAll("img")).filter((img) => {
-    // Filter out logos, avatars, icons
     const parentClass = img.parentElement?.className || "";
     const isLogo = parentClass.includes("logo") || img.alt.includes("logo");
     const isIcon = img.width < 100 || img.height < 100;
@@ -33,26 +53,14 @@ function addFigureNumbers() {
   let figureCount = 0;
 
   images.forEach((img) => {
-    // Get alt text
     const altText = img.getAttribute("alt") || "";
     if (!altText.trim()) {
       return;
     }
 
-    // Check if already has figcaption with figure number
-    const existingFigure = img.closest("figure");
-    if (existingFigure) {
-      const existingFigcaption = existingFigure.querySelector("figcaption");
-      if (existingFigcaption && existingFigcaption.textContent.match(/^(图|Figure)\d+:/)) {
-        figureCount++;
-        return;
-      }
-    }
-
-    // Increment counter
     figureCount++;
 
-    // Create figure wrapper if it doesn't exist
+    // Wrap in <figure> if not already wrapped
     let figure = img.closest("figure");
     if (!figure) {
       figure = document.createElement("figure");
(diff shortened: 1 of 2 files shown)


bhuvan-somisetty:
Hey @rootsongjc @mesutoezdil — submitted a fix for #699. This addresses the build-time changelog global state leak across locales as well as the client-side figure caption race condition during SPA navigation. Ready for review when you have a moment!

bhuvan-somisetty:
Updated imageFigureNumber.js to match root doc paths (e.g. /docs, /zh/docs) and preserve <a> wrappers when images are linked. Also synced the issue and PR descriptions to keep the checks aligned.

src/client/imageFigureNumber.js
```diff
@@ -2,28 +2,48 @@
...
-  setTimeout(() => addFigureNumbers(), 100);
-  setTimeout(() => addFigureNumbers(), 500);
-  setTimeout(() => addFigureNumbers(), 1000);
+  const locale = localeFromPathname(location.pathname);
+
+  // A single rAF tick is sufficient: Docusaurus has already committed the new
+  // page content to the DOM by the time onRouteDidUpdate fires. Using rAF
+  // instead of multiple staggered setTimeouts avoids duplicate caption
+  // injections when routes are navigated quickly.
+  requestAnimationFrame(() => addFigureNumbers(locale));
```
mesutoezdil, line 30: img.width is 0 before a lazy image loads, so the icon filter skips it. the old staggered timeouts hid that. one raf tick may miss slow images. maybe re-run on img load or use naturalWidth.

src/plugins/changelog/index.js
```diff
@@ -94,7 +95,7 @@ function processSection(section) {
   publishTimes.add(`${date}T${hour}:00`);
 
   return {
-    title: title.replace(/ \(.*\)/, ''),
+    title: title.replace(/ \(.*\)/, ""),
```
mesutoezdil, line 98: unrelated quote change. file uses single quotes.

mesutoezdil (commented):
good find on both bugs. note: #682 touches the same figure code, coordinate merge order.

src/client/imageFigureNumber.js
mesutoezdil, line 29: checked this on the live site and narrowed it: markdown images carry width and height, so those are fine. but raw tags like the `` on developers/protocol have no height, so img.height is 0 at rAF time and they get skipped as icons. the 500ms pass used to catch them.

mesutoezdil: still open on the current head: `docs/developers/protocol.md` on master carries `` with no height, so img.height is 0 at rAF time and that figure is filtered out as an icon.

bhuvan-somisetty:
Thanks @mesutoezdil! Pushed an update addressing both notes:

- Reverted the quote change in `src/plugins/changelog/index.js` back to single quotes.
- Updated `isIconImage` in `src/client/imageFigureNumber.js` to inspect explicit `width`/`height` attributes (such as `` on `docs/developers/protocol.md`) so images aren't filtered out when `height` is 0 at initial rAF time. Also added a `load` listener pass for slow/lazy loading images.

bhuvan-somisetty:
Updated the image `load` event listener callback in `src/client/imageFigureNumber.js` to dynamically evaluate `localeFromPathname(window.location.pathname)` when the image finishes loading. This ensures figure captions always match the active route locale even if the user navigates across pages while images are loading.

src/client/imageFigureNumber.js
```diff
@@ -2,88 +2,148 @@
...
+    // Re-run captioning when lazy/slow images complete loading
+    if (!img.complete && !img.dataset.hasLoadListener) {
+      img.dataset.hasLoadListener = "true";
+      img.addEventListener(
+        "load",
+        () => addFigureNumbers(localeFromPathname(window.location.pathname)),
+        { once: true },
+      );
+    }
+    return !isIconImage(img);
```
mesutoezdil: an unloaded image with no width/height attrs defaults to not-icon here, so small icons get captioned before loading. once isIconImage flips true after load, is the earlier figure/figcaption ever removed?

bhuvan-somisetty: good catch, fixed - now unwraps the figure/figcaption if it turns out to be an icon after load

src/client/imageFigureNumber.js
```diff
@@ -2,88 +2,172 @@
...
-    const parentClass = img.parentElement?.className || "";
-    const isLogo = parentClass.includes("logo") || img.alt.includes("logo");
-    const isIcon = img.width < 100 || img.height < 100;
-    return !isLogo && !isIcon;
+    // Re-run captioning when lazy/slow images complete loading
+    if (!img.complete && !img.dataset.hasLoadListener) {
+      img.dataset.hasLoadListener = "true";
+      img.addEventListener(
+        "load",
+        () => addFigureNumbers(localeFromPathname(window.location.pathname)),
```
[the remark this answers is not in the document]
bhuvan-somisetty: good catch, fixed - the load callback now re-checks the route before adding captions

src/client/imageFigureNumber.js
mesutoezdil: naturalWidth plus the load listener covers this, thanks.

src/plugins/changelog/index.js
mesutoezdil: Quote style is reverted here, this looks resolved.

mesutoezdil (commented):
The link check flaked, it never emitted a payload.
Both locale builds passed, so a re-run should clear it.
#682 is closed now, merge-order coordination no longer applies.

mesutoezdil (approved):
LGTM. Both bugs check out and the fixes match the diagnosis:

- `publishTimes` and `authorsMap` are now created inside `loadContent()` and threaded through `processSection`, so the zh pass no longer inherits the en timestamps and the hour-decrement loop stops shifting Chinese changelog dates.
- Locale now comes from `location.pathname` instead of `document.documentElement.lang`, which removes the race with the async `lang` update, and the widened route regex covers the root `/docs` and `/zh/docs` paths.
- The icon heuristic uses `naturalWidth || clientWidth || width` plus a `load` listener, so lazy images that report width 0 on the first frame get re-checked. That was my one concern and it is handled.
- Parent `<a>` links are preserved when wrapping images.

CI is green after the re-run; the earlier failure was the link checker never emitting a payload, not a broken link. #682 is closed, so there is no merge-order conflict on the figure code.

PR #708, opened 2026-08-03
docs(mthreads): fix device memory unit to 512 MiB and clarify sgpu-memory

veyron-kairo:
## What this does

Fixes the device-memory unit and wording on the Moore Threads (mthreads) pages.

- The **enable** page said each `sgpu-memory` unit is `512 MB`, while the **specify device memory** page said `512 MiB`. The device plugin multiplies the requested value by 512, and the doc's own example (`32` → 16Gi) only works out in MiB (32 × 512 MiB = 16384 MiB = 16 GiB). So `512 MiB` is correct — this makes both pages consistent.
- Dropped the "allocate a **percentage** size of device memory" wording. The value isn't a percentage; it's a plain integer count of 512-MiB units. Reworded to match the style of the other device pages ("Each unit ... represents ...").
- Clarified the example comment to show the arithmetic (`32 units x 512 MiB = 16 GiB`).

## Scope

Applied to the current (`/docs/next`) docs and the `v2.9.0` released snapshot, in both English and Chinese, following the "fixing the latest stable version" workflow in `AGENTS.md` (edit `docs/` + `i18n/zh/.../current/`, then sync the changed files into `versioned_docs/version-v2.9.0/` and the zh `version-v2.9.0/` mirror). Older frozen version snapshots were left untouched.

## Checks

`npx prettier --check` and `npx markdownlint` pass on the changed files. The change is in-line prose only (no links or structure changed).

## AI assistance disclosure

used chat gpt to make the pr description more fluent and enhanced my writing.

## Summary by CodeRabbit

- **Documentation**
  - Clarified Mthreads GPU-sharing installation parameters and Helm settings.
  - Corrected `sgpu-memory` units from 512 MB to 512 MiB.
  - Explained that `mthreads.com/sgpu-memory` is an optional memory allocation resource.
  - Updated examples to show that 32 units provide 16 GiB of device memory.
  - Synchronized these corrections across English and Chinese documentation.

Issue #660, opened 2026-07-27
Documentation versioning depends on a manual process, allowing releases and docs to drift


Adii-45:
**Page URL**:
Repository-wide documentation versioning process (not limited to a single page)

**What is wrong**:

The website's documentation versioning currently depends on a manual process, which allows published documentation to fall behind HAMI releases!

While investigating the documentation workflow, I found that the repository does not contain any workflow that synchronizes documentation versions when a new HAMI release is published.. The only documentation-related workflow (`docs-health.yml`) validates links and documentation health, but it does not update versioned documentation!

The repository itself also documents a previous regression where a manual versioning step was missed, resulting in documentation remaining behind the latest release. This indicates that the current process relies on maintainers remembering to perform a manual step after each release!

Because of this, users can temporarily receive stale documentation even though a newer HAMI release is already available.

**Evidence**

- Only one documentation workflow exists (`.github/workflows/docs-health.yml`), and it performs documentation validation rather than release/version synchronization.
- `AGENTS.md` documents a previous incident where missing a manual versioning step caused documentation to lag behind a release.
- The version history shows that documentation snapshots are created manually after releases rather than automatically as part of the release process.

**What it should say instead**:

The documentation release/versioning process should ensure that published documentation remains in sync with HAMI releases. If the manual process is intentional, it should be clearly documented so contributors can consistently perform the required post-release steps..!

Adii-45:
Hi @fishman @rootsongjc!
I'm working on a fix for this and will put together a PR soon.. Feel free to assign it to me if that sounds good.. Feedback is appreciated!

rootsongjc:
Do you have a plan how to sync the docs across versions? @Adii-45

Adii-45:
Thanks for the quick response @rootsongjc! I have a preliminary idea, but I'd like to validate it against the current release/versioning workflow before implementing anything...

My current plan is to identify the point where a new HAMI release is published and ensure the website's versioned documentation stays in sync from there, while preserving the existing manual versioning flow if needed! I'll investigate the current release pipeline in more detail and share a concrete proposal for the PR before making/implementing any behavioral changes..!

Adii-45:
Hi @rootsongjc,

Yes - I have a preliminary plan!

[...]

PR #714, opened 2026-08-03
docs(device): warn against Kubernetes quantity suffixes for device memory requests

veyron-kairo:
## What this does

Adds a warning to every backend's device-memory doc about a real footgun that currently has **no documentation** and silently causes out-of-memory failures.

The per-device memory resources — `nvidia.com/gpumem`, `hygon.com/dcumem`, `cambricon.com/mlu.smlu.vmemory`, `iluvatar.ai/<card-type>.vMem` — each take a **plain integer**. If a user writes a Kubernetes quantity such as `16Gi` (a very natural thing to try for "16 GB of memory"), HAMi parses it as its byte count (~17 billion), which overflows HAMi's 32-bit memory field and is silently truncated, often to `0`. A zero memory request passes the scheduler's memory check unconditionally, so the pod is scheduled onto a device that is already full and then OOMs at runtime.

I ran into this while fixing the underlying overflow in the code ([Project-HAMi/HAMi#2285](https://github.com/Project-HAMi/HAMi/pull/2285) / [#2278](https://github.com/Project-HAMi/HAMi/issues/2278)): the device plugins now reject an overflowing value, but nothing tells users the correct format in the first place. This documents it.

## Changes

- Adds a `:::warning` to the nvidia, hygon, cambricon, and iluvatar memory docs explaining the integer format and the failure mode.
- Fixes hygon's opening line, which said "allocate a **percentage size** of device memory" while also saying each unit is 1 MiB — it uses fixed 1-MiB units, not a percentage (same wording bug as the mthreads page). Aligns the example comment.
- mthreads has the same footgun; its unit wording is already being fixed in #708, so this PR leaves it out to avoid a conflict — happy to add the same warning there as a follow-up.

## Scope

English + Chinese, current (`/docs/next`) and the `v2.9.0` released snapshot, following the "fixing the latest stable version" workflow in `AGENTS.md`. Syncing current → `v2.9.0` also carried two pre-existing minor fixes that current already had (the versioning drift noted in #660): `3000m`→`3000 MiB` and `can't`→`cannot` on the nvidia page, and a helm-flag spelling on the iluvatar page.

## Checks

`npx prettier --check` and `npx markdownlint` pass on the changed files.

## Notes

A native Chinese speaker is welcome to refine the zh phrasing of the new warnings.

## AI assistance disclosure

I used chat gpt to write the Chinese counter of this pr and read and understood everything translating it to English after writing .


mesutoezdil (commented):
good catch and clear warning. rebase so the docs health ci runs on this pr.

i18n/zh/docusaurus-plugin-content-docs/version-v2.9.0/userguide/iluvatar-device/enable-iluvatar-gpu-sharing.md
```diff
@@ -30,7 +30,7 @@ title: 启用天数智芯 GPU 共享
 - 安装 HAMi 时设置 `devices.iluvatar.enabled=true`
 
 ```bash
-helm install hami hami-charts/hami --set scheduler.kubeScheduler.imageTag={your kubernetes version} --set devices.iluvatar.enabled=true -n kube-system
+helm install hami hami-charts/hami --set scheduler.kubeScheduler.image.tag={your kubernetes version} --set devices.iluvatar.enabled=true -n kube-system
```
mesutoezdil, line 33: same question as in #708: does the v2.9 chart use image.tag? if it still uses imageTag this breaks the snapshot.

docs/userguide/nvidia-device/specify-device-memory-usage.md
```diff
@@ -12,6 +12,12 @@ resources:
     nvidia.com/gpumem: 3000 # Each GPU contains 3000 MiB device memory
 ```
 
+:::warning
```
mesutoezdil, line 15: your own hami pr #2285 makes the plugin reject an overflowing value instead of truncating it. from which hami release does this page describe the old behaviour?

docs/userguide/nvidia-device/specify-device-memory-usage.md
```diff
@@ -12,6 +12,12 @@ resources:
...
+
+Set `nvidia.com/gpumem` as a **plain integer** (a count of MiB), not a Kubernetes quantity. A suffixed value such as `16Gi` is parsed as its byte count (about 17 billion), which overflows HAMi's 32-bit memory field and is silently truncated, often to `0`. A zero memory request passes the scheduler's memory check unconditionally, so the pod can be placed on a GPU that is already full and then fails with out-of-memory errors at runtime. Use the integer form shown above (for example `3000`).
+
```
mesutoezdil, line 18: nvidia.com/gpumem-percentage and nvidia.com/gpucores just below take a plain integer too. do they hit the same field, and should the warning cover them?

docs/userguide/nvidia-device/specify-device-memory-usage.md
```diff
@@ -12,6 +12,12 @@ resources:
     nvidia.com/gpumem: 3000 # Each GPU contains 3000 MiB device memory
 ```
 
+:::warning
+
+Set `nvidia.com/gpumem` as a **plain integer** (a count of MiB), not a Kubernetes quantity. A suffixed value such as `16Gi` is parsed as its byte count (about 17 billion), which overflows HAMi's 32-bit memory field and is silently truncated, often to `0`. A zero memory request passes the scheduler's memory check unconditionally, so the pod can be placed on a GPU that is already full and then fails with out-of-memory errors at runtime. Use the integer form shown above (for example `3000`).
```
mesutoezdil, line 17: confirmed the mechanism: with Memreq 0 and MemPercentagereq 101, memreq stays 0 at nvidia/device.go:918 and the check at 927 then passes on any card. the source even carries a `//This incurs an issue` comment there.

docs/userguide/iluvatar-device/enable-iluvatar-gpu-sharing.md
```diff
@@ -121,6 +121,12 @@ Each unit of `iluvatar.ai/<card-type>.vMem` represents 256 MB of device memory.
 
 :::
 
+:::warning
+
+Set `iluvatar.ai/<card-type>.vMem` as a **plain integer** (a count of 256 MB units), not a Kubernetes quantity. A suffixed value such as `16Gi` is parsed as its byte count (about 17 billion), which overflows HAMi's 32-bit memory field and is silently truncated, often to `0`. A zero memory request passes the scheduler's memory check unconditionally, so the pod can be placed on a GPU that is already full and then fails with out-of-memory errors at runtime. Use the integer form shown above (for example `64`).
```
mesutoezdil, line 126: hami #2285 is titled reject int32-overflowing memory requests in iluvatar and mthreads, so on iluvatar this now errors instead of truncating. should the iluvatar wording differ from the nvidia, hygon and cambricon ones?

mesutoezdil:

mesutoezdil:

PR #715, opened 2026-08-04
fix: patch brace-expansion, fast-uri, and postcss CVEs disclosed 2026-08-03

mesutoezdil:
/kind bug

Three new advisories were published on 2026-08-03 against dependencies already pinned in the lockfile:

- **brace-expansion** (CVE-2026-69152, GHSA-rgw5-rvv9-x895): DoS via unbounded intermediate arrays. This bypasses the CVE-2026-14257 mitigation that the existing `5.0.8` override was pinned to. Affected `< 5.0.9`.
- **fast-uri** (CVE-2026-18446): host confusion via a backslash authority introducer lets an attacker bypass host allowlist / SSRF checks that rely on fast-uri's parsing. Affected `3.0.0 - <3.1.5`; pulled in transitively via `ajv`.
- **postcss** (GHSA-fxqj-rqcc-2cmp): an incomplete fix of GHSA-6g55-p6wh-862q. `PreviousMap.loadFile()` can still be made to read an arbitrary source-map file when `from` is unset. Affected `<=8.5.22`.

Fix: bump the existing `brace-expansion` override to the patched `5.0.9`, and add `fast-uri` and `postcss` overrides pinned to their patched versions, following the same pattern already used for `brace-expansion`, `markdownlint-cli`, etc.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,7 +34,7 @@
         "swc-loader": "^0.2.7"
       },
       "engines": {
-        "node": ">=20.18.1"
+        "node": ">=20.19.0"
       }
     },
     "node_modules/@11ty/gray-matter": {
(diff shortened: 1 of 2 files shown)


rootsongjc (approved):
/lgtm

Issue #689, opened 2026-07-31
RFC: Reorganize the HAMi docs. Which one should I use: HAMi, HAMi-DRA, Volcano vGPU, or KAI Scheduler?


archlitchi:
We're reorganizing [hami-project.io](https://project-hami.io). Before we rebuild it, tell us what's broken.

Full plan: [Google Doc](https://docs.google.com/document/d/1SrNv4fU30VJie80k1RzY79ug_s8Ll6vdAR3jkgUTFX4/edit?tab=t.0)

**3 problems we keep hearing:**

1. **Hard to navigate.** The docs are a flat list. New users don't know where to start.
2. **HAMi or HAMi-DRA?** Both are documented, but nothing says when to pick which.
3. **Capabilities don't match.** HAMi / HAMi-DRA / Volcano vGPU / KAI Scheduler don't do the same things. Users only find out in production.

**What we'll build (react to each):**

1. Guided nav: Get started → Concepts → Install → Pick your approach → Vendor guides → Integrations → Operate
2. One "HAMi vs HAMi-DRA" decision page
3. A capability matrix showing who supports what

**Answer any of these (one line is fine):**

1. Which doc do you open first? Which is hardest to find?
2. Why did you pick HAMi or HAMi-DRA?
3. Do you run HAMi with Volcano or KAI? What's missing from those guides?
4. If we add a "choose your path" flow, what should it ask first?

We're targeting completion by **Aug 14, 2026**. Comment by **Aug 14**. Or chat in [Slack](https://cloud-native.slack.com/archives/C07T10BU4R2) / [Discord](https://discord.gg/Amhy7XmbNq).

/area docs
/kind documentation

archlitchi:
CC @mesutoezdil @fishman @rootsongjc

singhsrijan46:
@archlitchi 
I would like to work on this issue. Please assign it to me.

fishman:
@singhsrijan46 you can suggest changes or submit a PR.

@archlitchi  I agree, it's "crowded" as they say

WaliTheOrchestrator:
I am currently working on my proposal. I will try my best to present it in upcoming english meeting on wednesday

AyushSrivastava1818:
The missing piece isn't any single doc — it's the "which one do I even use" page
that should exist before the installation guides.

Two things a comparison page should say plainly, up front:
1. **HAMi vs HAMi-DRA**: pick classic HAMi unless you're on K8s ≥1.34 with the
[...]

PR #720, opened 2026-08-04
test(docs): validate the Kubernetes manifest examples parse

veyron-kairo:
## What this does

Adds a documentation test that validates the Kubernetes manifest examples in the docs actually parse, and fixes one that didn't.

Many pages show YAML the reader is expected to copy and `kubectl apply`. Nothing checked that those examples are valid, so a broken one could ship. This adds a `node --test` check (`test/doc-manifests.test.mjs`) that:

- walks `docs/`, `tutorials/`, and the zh `current/` docs;
- takes the fenced `yaml` blocks that declare both `apiVersion` and `kind` — i.e. real manifests, not bare `resources:` snippets or the annotated example lists that deliberately repeat keys;
- asserts each one parses.

Running it surfaced one broken example: the Chinese metax scheduling guide wrote the pod's `annotations` key and its value on a single line (`annotations: hami.io/node-scheduler-policy: "spread"`), which is invalid YAML. I fixed it to the nested form the English page already uses, in both `current` and the `v2.9.0` snapshot (per the versioning workflow in `AGENTS.md`). Fixing the block also let Prettier format it, which added a space in a `command: ["sleep", "infinity"]` array on the same block.

Adds `js-yaml` as a devDependency for the check.

## Checks

`npm test`, `npm run lint`, and `npm run format:check` pass. I confirmed the check fails (with a clear `file (manifest #N): reason`) when the manifest is left broken, so it guards against regressions rather than just passing today.

## AI assistance disclosure

I used AI assistance (Claude Code) while locating the broken example. I reviewed and verified it, confirmed it catches the regression, ran the checks locally, and take responsibility for the change.

## Summary by CodeRabbit

* **Documentation**
  * Corrected MetaX GPU scheduling examples in the Chinese documentation.
  * Fixed YAML annotation nesting and formatting while preserving the existing `spread` scheduling behavior.
  * Updated both current and v2.9.0 documentation examples for consistency.

* **Tests**
  * Added automated validation for YAML blocks in current documentation, helping detect malformed Kubernetes examples before publication.

### DIFF
--- a/test/doc-manifests.test.mjs
+++ b/test/doc-manifests.test.mjs
@@ -1,0 +1,61 @@
+/**
+ * Validates the Kubernetes manifest examples embedded in the documentation.
+ *
+ * Many pages show YAML the reader is expected to copy and apply. A block is
+ * treated as a manifest only when it declares both `apiVersion` and `kind`;
+ * bare snippets (a lone `resources:` block) and annotated example lists (which
+ * deliberately repeat keys to contrast valid and invalid requests) are not full
+ * manifests and are skipped. Every manifest that remains must parse as YAML, so
+ * a broken example can't ship.
+ *
+ * Versioned snapshots under versioned_docs/ are frozen and excluded, matching
+ * the lint:md scope.
+ */
+import { describe, it } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync, readdirSync, statSync } from "node:fs";
+import { join } from "node:path";
+
+import { loadAll } from "js-yaml";
+
+const roots = ["docs", "tutorials", "i18n/zh/docusaurus-plugin-content-docs/current"];
+
+function markdownFiles(dir, out = []) {
+  for (const entry of readdirSync(dir)) {
+    const p = join(dir, entry);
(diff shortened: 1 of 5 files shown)


test/doc-manifests.test.mjs
```diff
@@ -0,0 +1,63 @@
...
+    if (entry.isDirectory()) markdownFiles(p, out);
+    else if (entry.isFile() && (entry.name.endsWith(".md") || entry.name.endsWith(".mdx")))
+      out.push(p);
+  }
+  return out;
+}
+
+function manifestBlocks(markdown) {
+  const blocks = [];
+  const fence = /```ya?ml\n([\s\S]*?)```/g;
```
mesutoezdil, line 36: fences with meta like ```yaml title="x" do not match this regex, those blocks are silently skipped. allow text after the language tag.

mesutoezdil (commented):
useful check and a real fix. rebase so the docs health ci runs on this pr.

mesutoezdil:
Closing this for now: new CI checks were not requested and add maintenance cost while the docs are being reorganized in #689.

PR #724, opened 2026-08-05
fix: hoist `useBaseUrl` calls to follow Rules of Hooks

### LINKED ISSUE #723: `useBaseUrl` breaks Rules of Hooks in a few components
`useBaseUrl` is a React hook, but a few spots call it inside conditionals, `.map()`, or after an early return:

- `src/pages/index.js` : inside ecosystem / vendor maps
- `src/pages/case-studies.js` :  inside `caseStudies.map`
- `src/theme/BlogPostItem/Header/index.js` :  only when `cover` is set
- `src/theme/Footer/index.js` :  `useEffect` / `useBaseUrl` after `if (!footer) return null`

No ESLint in the repo (`npm run lint` is markdown only), so nothing catches it. Would be good to fix the call sites and add `eslint-plugin-react-hooks`.

Anshumancanrock:
**What type of PR is this?**

/kind bug

`useBaseUrl` is a React hook, but we were calling it inside `.map()`s, conditionals, and after an early return in Footer. It's against the Rules of Hooks and will blow up if that data becomes conditional.

Fix: use `useBaseUrlUtils()` / hoist the hook calls to the top level.

Fixes #723

**Checklist**:

- [x] Chinese translation updated if English docs changed (or noted why not) — N/A, JS only

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL handling across homepage, case studies, blog covers, and footer links.
  * Ensured images and navigation links resolve correctly in different site paths and locales.
  * Improved footer rendering consistency when footer configuration is unavailable.

### DIFF
--- a/src/pages/case-studies.js
+++ b/src/pages/case-studies.js
@@ -1,6 +1,6 @@
 import Layout from "@theme/Layout";
 import useDocusaurusContext from "@docusaurus/useDocusaurusContext";
-import useBaseUrl from "@docusaurus/useBaseUrl";
+import { useBaseUrlUtils } from "@docusaurus/useBaseUrl";
 import caseStudies from "../data/caseStudies";
 import styles from "./case-studies.module.css";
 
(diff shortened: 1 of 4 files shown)


Creativeklvn:
The changes correctly move the hooks out of loops/conditional paths while keeping the existing URL behavior.

rootsongjc (approved):
/LGTM

PR #732, opened 2026-08-06
fix(edge): nested lists and attribute > leaks in markdown negotiation

### LINKED ISSUE #731: Nested lists get flattened / `>` in attributes leaks into markdown
Follow-ups from #688. Not fixed there.

### 1. Nested lists get flattened

`htmlToMarkdown` matches list items with a non-greedy `*?`, so the first `</li>` wins. Nested items get promoted to the top level and the numbering goes wrong.

Repro HTML:

```html
<ol>
  <li>Outer step one
    <ul>
      <li>sub A</li>
      <li>sub B</li>
    </ul>
  </li>
  <li>Outer step two</li>
</ol>
```

What comes out:

```
1. Outer step one
- sub A
2. sub B
3. Outer step two
```

Hits the guides that nest lists under ordered steps (`gpu-virtualization`, `enable-cambricon-mlu-sharing`, `enable-metax-gpu-schedule`, `grafana-dashboard`, `hami-webui-development-guide`, and their built variants). On the WebUI guide, sub-bullets come out as steps 2–5 and the real step 2 lands as step 6.

Only affects `Accept: text/markdown` clients, but for those the procedure text is just wrong.

### 2. Attribute values containing `>` leak

Tag scanning uses `[^>]*`, so a literal `>` inside a quoted attribute ends the match early:

```html
<span title="a > b">Hello</span>
```

→ `b">Hello`

Docusaurus usually emits `&gt;` instead, so this mostly doesn't fire on the live site today. Still a real hole in the parser.

### Fix

Depth-aware list walking for (1), quote-aware tag matching for (2), plus regression tests next to the ones.

Happy to take both.

Anshumancanrock:
**What type of PR is this?**

/kind bug

Fixes the two follow-ups from #688.

Nested lists were getting flattened because `<li>.*?</li>` closed at the first nested `</li>`, so sub-items got promoted and parent steps renumbered. Walk `ul`/`ol`/`li` by depth instead, and unwrap `<p>` before list conversion so Docusaurus-style items don't come out as `1.\n\ntext`.

Quoted `>` in attributes was leaking into the body (`title="a > b"` → `b">…`) because tag scans use `[^>]*`. Mask those before the scans, restore after.

Also keep leading indent when collapsing spaces, otherwise nested list markers get wiped.

Fixes #731

**Checklist**:

- [x] Chinese translation updated if English docs changed (or noted why not) (n/a, no docs touched)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Markdown conversion for nested ordered and unordered lists, multiline items, and incomplete lists.
  - Ordered lists now honor their starting number.
  - Prevented `>` characters inside quoted HTML attributes from disrupting conversion.
  - Preserved indentation more reliably and left stray `<` characters in text unchanged.

- **Tests**
  - Added regression coverage for nested lists, ordered-list numbering, protected angle characters, encoded values, and stray text symbols.

### DIFF
--- a/netlify/edge-functions/markdown-negotiation.js
+++ b/netlify/edge-functions/markdown-negotiation.js
@@ -314,11 +314,13 @@ function toMarkdownTable(tableHtml) {
 }
 
 function htmlToMarkdown(html, pageUrl) {
-  const rawTitle = html.match(/<title[^>]*>([\s\S]*?)<\/title>/i)?.[1] ?? "";
-  const rawDescription =
-    html.match(/<meta[^>]+name=["']description["'][^>]+content=["']([^"']*)["'][^>]*>/i)?.[1] ?? "";
+  html = protectQuotedAngles(html);
+  const rawTitle = restoreQuotedAngles(html.match(/<title[^>]*>([\s\S]*?)<\/title>/i)?.[1] ?? "");
+  const rawDescription = restoreQuotedAngles(
+    html.match(/<meta[^>]+name=["']description["'][^>]+content=["']([^"']*)["'][^>]*>/i)?.[1] ?? "",
+  );
 
-  let content = protectQuotedAngles(html.match(/<main[^>]*>([\s\S]*?)<\/main>/i)?.[1] ?? html);
+  let content = html.match(/<main[^>]*>([\s\S]*?)<\/main>/i)?.[1] ?? html;
 
   content = content
     .replace(/<script[\s\S]*?<\/script>/gi, "")
--- a/netlify/edge-functions/markdown-negotiation.js
+++ b/netlify/edge-functions/markdown-negotiation.js
@@ -83,6 +83,53 @@ function createBlockStore() {
   };
 }
 
+// Same idea as the block-store marker: hide `>` inside quoted attrs so `[^>]*`
+// tag scans don't stop early.
+const ATTR_GT = "\u0001";
+
+function findTagClose(html, openIndex) {
+  let quote = null;
+  for (let i = openIndex + 1; i < html.length; i++) {
+    const c = html[i];
+    if (quote) {
+      if (c === quote) {
+        quote = null;
+      }
+      continue;
+    }
+    if (c === '"' || c === "'") {
+      quote = c;
+    } else if (c === ">") {
+      return i;
+    }
+  }
+  return -1;
+}
+
+function protectQuotedAngles(html) {
+  let result = "";
+  let cursor = 0;
+  for (let i = 0; i < html.length; i++) {
+    if (html[i] !== "<") {
+      continue;
+    }
+    const close = findTagClose(html, i);
+    if (close === -1) {
+      break;
+    }
+    result += html.slice(cursor, i);
+    result += html.slice(i, close).replace(/>/g, ATTR_GT);
+    result += ">";
+    cursor = close + 1;
+    i = close;
+  }
+  return result + html.slice(cursor);
+}
+
+function restoreQuotedAngles(text) {
+  return text.includes(ATTR_GT) ? text.replace(/\u0001/g, ">") : text;
+}
+
 // Depth-tracked because the TOC containers nest divs, so the first closing tag
 // isn't the matching one.
 function dropElements(html, tagName, classPattern) {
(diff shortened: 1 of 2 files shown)


netlify/edge-functions/markdown-negotiation.js
```diff
@@ -83,6 +83,53 @@ function createBlockStore() {
   };
 }
 
+// Same idea as the block-store marker: hide `>` inside quoted attrs so `[^>]*`
+// tag scans don't stop early.
+const ATTR_GT = "\u0001";
+
+function findTagClose(html, openIndex) {
```
mesutoezdil, line 90: a stray unescaped < in text opens a fake tag here and can mask real tags up to the next >. docusaurus should always emit &lt;, so likely fine, just confirming you considered it.

mesutoezdil (commented):
solid fix, good tests. ci build is green even though the checklist box is empty.

netlify/edge-functions/markdown-negotiation.js
```diff
@@ -113,6 +160,105 @@ function dropElements(html, tagName, classPattern) {
...
+}
+
+function renderList(inner, ordered) {
+  const items = extractDirectLis(inner);
+  if (items.length === 0) {
+    return "";
+  }
+  return items
+    .map((item, index) => {
+      const marker = ordered ? `${index + 1}. ` : `- `;
```
mesutoezdil, line 234: ol start="3" is ignored, numbering restarts at 1. rare in docs but easy to read from the open tag.

netlify/edge-functions/markdown-negotiation.js
```diff
@@ -83,6 +83,53 @@ function createBlockStore() {
   };
 }
 
+// Same idea as the block-store marker: hide `>` inside quoted attrs so `[^>]*`
+// tag scans don't stop early.
+const ATTR_GT = "\u0001";
```
mesutoezdil, line 88: if the page body ever contains a literal \u0001 it gets turned into >. a longer private marker would be safer.

test/markdown-negotiation.test.mjs
```diff
@@ -257,6 +257,41 @@ describe("block structure", () => {
...
+    assert.equal(
+      await body(
+        "<main><ol><li><p>Install deps</p><ul><li>Node 20</li><li>npm ci</li></ul></li>" +
+          "<li><p>Build the site</p></li></ol></main>",
+      ),
+      "1. Install deps\n   - Node 20\n   - npm ci\n2. Build the site",
+    );
+  });
+
+  it("keeps nested ordered lists nested", async () => {
```
mesutoezdil, line 281: no test for a three level list. the recursion should handle it, one case would lock it in.

netlify/edge-functions/markdown-negotiation.js
Anshumancanrock: done, we read start off the open tag now and number from there. covered with a test.

netlify/edge-functions/markdown-negotiation.js
Anshumancanrock: fair point. switched to a longer private-use marker (\uE000ATTRGT\uE000).

test/markdown-negotiation.test.mjs
Anshumancanrock: added one. also fixed a small indent bug `trimStart` was wiping on continuation lines past two levels.

Anshumancanrock:
hii @mesutoezdil , please take a look now. Thanks!

netlify/edge-functions/markdown-negotiation.js
```diff
@@ -168,9 +324,11 @@ function toMarkdownTable(tableHtml) {
 }
 
 function htmlToMarkdown(html, pageUrl) {
-  const rawTitle = html.match(/<title[^>]*>([\s\S]*?)<\/title>/i)?.[1] ?? "";
-  const rawDescription =
-    html.match(/<meta[^>]+name=["']description["'][^>]+content=["']([^"']*)["'][^>]*>/i)?.[1] ?? "";
+  html = protectQuotedAngles(html);
+  const rawTitle = restoreQuotedAngles(html.match(/<title[^>]*>([\s\S]*?)<\/title>/i)?.[1] ?? "");
```
mesutoezdil, line 328: the new tests cover the body but not this path. is there a case for a title or meta description whose attribute contains a `>`?

netlify/edge-functions/markdown-negotiation.js
```diff
@@ -83,6 +83,57 @@ function createBlockStore() {
...
+    if (c === '"' || c === "'") {
+      quote = c;
+    } else if (c === ">") {
+      return i;
+    }
+  }
+  return -1;
+}
+
+function protectQuotedAngles(html) {
```
mesutoezdil, line 108: this walks every character of the page html on each request, inside an edge function. did you measure the added time on a large docs page?

netlify/edge-functions/markdown-negotiation.js
```diff
@@ -113,6 +164,111 @@ function dropElements(html, tagName, classPattern) {
...
+function convertLists(html) {
+  const open = /<(ul|ol)\b[^>]*>/gi;
+  let result = "";
+  let cursor = 0;
+  let match;
+
+  while ((match = open.exec(html))) {
+    result += html.slice(cursor, match.index);
+    const ordered = match[1].toLowerCase() === "ol";
+    const startAttr = ordered ? match[0].match(/\bstart\s*=\s*["']?(-?\d+)/i)?.[1] : null;
```
mesutoezdil, line 253: the `-?` accepts a negative start, and `<ol start="-2">` then emits `-2. a`, which markdown renders as text rather than a list. i doubt docusaurus ever emits it, but is the sign wanted here?

netlify/edge-functions/markdown-negotiation.js
mesutoezdil, line 328: answering my own two earlier questions: i ran the branch. a `>` inside a title or description attribute comes out right, and 378 KB of html takes about 7 ms, so neither is a problem. the 79 tests pass.

mesutoezdil:

PR #754, opened 2026-08-11
docs(core-concepts): clarify which cu/nvml symbols libvgpu.so intercepts

ipsitapp8:
**What type of PR is this?**

/kind documentation

The GPU virtualization page says HAMi-Core hijacks NVIDIA library functions "starting with `cu`/`nvml`", and that it intercepts "all function calls starting with `cu` and `nvml`". That is broader than what `libvgpu.so` does.

`dlsym` is overridden in `src/libvgpu.c`, but what comes back is decided by a lookup, not by the prefix:

- `nvml*` goes to `__dlsym_hook_section_nvml()`, which is an explicit list of `DLSYM_HOOK_FUNC(...)` entries. A symbol that is not in the list returns NULL and falls through to the real `dlsym`.
- `cu*` is resolved against `libvgpu.so` itself first. If HAMi-Core defines no wrapper for it, it falls through the same way. The `cuGetProcAddress` path goes through `__dlsym_hook_section()`, which is the same kind of list.
- One family is skipped outright, in `find_symbols_in_table()` in `src/cuda/hook.c`:

  ```c
  /* Skip CUDA graph functions: let them fall through to real driver */
  if (strncmp(symbol, "cuGraph", 7) == 0) {
      return NULL;
  }
  ```

This matters when the page is used to debug an isolation gap. "Every `cu`/`nvml` call is intercepted" and "the calls in the hook table are intercepted" point at different causes for an allocation that got past the limit.

The change touches the `HAMi-Core` row in the component table and the `dlsym` paragraph in the interception flow section, in English and Chinese. Text only, no links or page structure changed.

AI assistance: I used Claude Code to check the existing wording against the HAMi-core source.

None. 4 changed lines, below the 100 line threshold that requires an issue first.

**Checklist**:

Note on the first box: markdownlint is clean across the repo and Prettier is clean on both changed files. Running `npm run format:check` on a Windows checkout reports every file in the repo, including untouched ones on master, because `core.autocrlf` gives CRLF line endings locally. That is a local artifact, not something in this diff.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that only explicitly registered NVIDIA functions are intercepted by the virtualization library.
  * Documented that unregistered functions resolve normally to the underlying driver.
  * Updated both English and Chinese GPU virtualization documentation.


rootsongjc (approved):
/lgtm

PR #778, opened 2026-08-15
fix(events): remove stale manual landing pages superseded by #701

Adi-xt:
# PR: Remove stale manual landing pages superseded by #701

## Title
fix(events): remove stale manual landing pages causing duplicate route warnings

### Problem
Running `npm run build` emits duplicate route warnings for two event landing pages, in both `en` and `zh` locales:

```
[WARNING] Duplicate routes found!
- Attempting to create page at /landing/coscup-2026, but a page already exists at this route.
- Attempting to create page at /landing/opensource-summit-korea, but a page already exists at this route.
This could lead to non-deterministic routing behavior.
```

### Root cause
`src/plugins/events/index.js` iterates over `landingEvents` and programmatically registers a route for each event at `/landing/${event.slug}`, rendering `EventLanding.js` with the event's `slug` prop.

However, `src/pages/landing/coscup-2026.js` and `src/pages/landing/opensource-summit-korea.js` are manual page files that render the exact same component with the exact same prop:

```jsx
import EventLanding from "@site/src/components/EventLanding";
export default () => <EventLanding slug="coscup-2026" />;
```

Since Docusaurus auto-generates routes for any file under `src/pages/`, this creates a route twice — once from the filesystem router, once from the plugin.

This is the same class of issue fixed in #701 ("generalize event landing page routes from data"), which removed two other stale manual pages (`kcd-vietnam.js`, `kubecon-japan.js`) for the identical reason but missed these two.

### Fix
Removed the two redundant manual page files:
- `src/pages/landing/coscup-2026.js`
- `src/pages/landing/opensource-summit-korea.js`

The plugin already handles routing/rendering for these events via `src/data/events.js`.

### Verification
- `npm run build` — duplicate route warnings no longer appear for either `en` or `zh` locale; both builds compile and generate successfully.
- `npm run serve` — manually visited `/landing/coscup-2026`, `/landing/opensource-summit-korea`, and their `/zh/` equivalents; all four render correctly via the plugin-generated route.

## Summary by CodeRabbit

* **Removed Pages**
  * Removed the COSCUP 2026 landing page.
  * Removed the Open Source Summit Korea landing page.

### DIFF
--- a/src/pages/landing/coscup-2026.js
+++ b/src/pages/landing/coscup-2026.js
@@ -1,3 +1,0 @@
-import EventLanding from "@site/src/components/EventLanding";
-
-export default () => <EventLanding slug="coscup-2026" />;
--- a/src/pages/landing/opensource-summit-korea.js
+++ b/src/pages/landing/opensource-summit-korea.js
@@ -1,3 +1,0 @@
-import EventLanding from "@site/src/components/EventLanding";
-
-export default () => <EventLanding slug="opensource-summit-korea" />;


src/pages/landing/coscup-2026.js
```diff
@@ -1,3 +0,0 @@
-import EventLanding from "@site/src/components/EventLanding";
-
```
mesutoezdil: verified: the plugin route covers this slug from events.js line 108, safe to delete.

mesutoezdil (commented):
fix is correct: both slugs are in src/data/events.js and only index.js remains under src/pages/landing, matching what #701 did. blockers: dco fails, the commit has no sign off, amend with git commit -s and force push. the docs health ci did not run, rebase to trigger it. and please use the pr template, kind and checklist are missing.

src/pages/landing/coscup-2026.js
Adi-xt: Thanks for the review! I’ve fixed the DCO issue by amending the commit with the required `Signed-off-by` line and force-pushed the updated commit.

I’ll also rebase onto the latest `master` to trigger the docs health CI, and update the PR description to follow the repository template with the required kind and checklist sections.

Adi-xt:
Thanks for the review! I’ve addressed the DCO requirement and rebased locally as requested. During the rebase, Git detected that the same fix is now already present in `master` via `c151fd30`, so my commit was skipped as an already-applied change.

Since the fix was independently incorporated into `master`, would you prefer that I keep this PR open for attribution, or close it and work on another issue?

mesutoezdil:
closing this: commit c151fd30 already removed both files and they are not on master any more, so this pr would be a no-op.

PR #767, opened 2026-08-12
docs(troubleshooting): add scheduler filter reason code reference

ipsitapp8:
## What

Adds a troubleshooting page for the reason codes the HAMi scheduler puts in `FilteringFailed` events when a Pod won't schedule.

## Why

When a Pod that requests HAMi resources stays Pending, the scheduler has already said why:

```plaintext
Warning  FilteringFailed  16s  hami-scheduler  2 nodes CardInsufficientMemory(node-a,node-b)
```

There are 17 of these codes in `pkg/device/common/common.go` and the site documents none of them. The closest page is `developers/scheduler-event-log.md`, but that one is a design proposal written in future tense and it only lists 7. So if you hit `CardTimeSlicingExhausted` or `AllocatedCardsInsufficientRequest`, there is nowhere to look it up.

## What's in the page

* All 17 codes with what triggers each one and how to fix it, grouped by the stage where the scheduler rejects the card.
* How to read both formats: the event (`N nodes <Code>(node-a,node-b)`) and the scheduler log line (`NodeUnfitPod ... reason="3/8 CardInsufficientMemory"`).
* Three things that are easy to get wrong. One event is emitted per reason code, so there is usually more than one to read. The events only appear when no node fits at all. And `AllocatedCardsInsufficientRequest` counts the cards that fit, not the ones that didn't.
* Turning on `-v=5` for per-device detail, how much log volume that costs, and how to revert.

## Notes

Everything was checked against HAMi v2.9.0 source and the page says which version it reflects. Added as a new file instead of editing `troubleshooting.md` so it doesn't clash with #728. Chinese translation included. Not backported to `versioned_docs/` since it's new content.

**Which issue(s) this PR fixes**:

N/A

**Checklist**:

- [x] `npm run lint` and `npm run format:check` pass
- [x] `npm run build` succeeds for both `en` and `zh`
- [x] Chinese translation updated if English docs changed
- [x] Commits are signed off (`git commit -s`)

**AI assistance disclosure**: This page's content and description were drafted with Claude Code from the HAMi v2.9.0 source and verified manually against it. I checked every reason code and link myself before opening the PR.

PR #741, opened 2026-08-08
docs(troubleshooting): add soft memory enforcement mechanics & libvgp…

ipsitapp8:
**What type of PR is this?**

/kind documentation

**What this PR does / why we need it**:

Adds a comprehensive technical guide and diagnostic runbook explaining how HAMi's soft CUDA-level memory enforcement works, why it can be bypassed, and how to debug each bypass scenario step by step.

Changes:
- **[NEW] `docs/troubleshooting/cuda-memory-enforcement.md`**: Explains the `libvgpu.so` interception chain, a soft vs hard enforcement comparison matrix, 5 documented bypass scenarios with diagnostic commands, and a 7-step quick checklist.
- **[MODIFY] `docs/troubleshooting/troubleshooting.md`** (en + zh): Adds a cross-reference tip pointing to the deep-dive page.
- **[MODIFY] `sidebars.js`**: Converts the `troubleshooting` single doc into a category with both pages.
- **[MODIFY] `i18n/zh/.../current.json`**: Adds the missing zh label for the new "Troubleshooting" sidebar category (previously showed in English).
- **[NEW] `i18n/zh/.../cuda-memory-enforcement.md`**: Chinese placeholder showing the English content with a "not yet translated" note, plus corrected FAQ anchor links (pointing at the actual Chinese-slugified anchor ids, not the English ones).

Note on the zh placeholder: a reviewer suggested dropping it since Docusaurus falls back to English automatically when a translated doc is missing. I tried that, but this site's docs plugin only applies that fallback at the routing/sidebar level, not when resolving relative markdown links from other zh files (or from this doc's own links, when it's reused as the zh fallback) — deleting it broke MDX compilation for two links. Verified locally with a full `en`+`zh` production build; keeping the placeholder is the only configuration that builds clean.

**Which issue(s) this PR fixes**:

Related to #656 ([LFX Mentorship] Expand HAMi GPU-sharing tutorials, labs, and troubleshooting guides)

**Checklist**:

- [x] `npm run lint` and `npm run format:check` pass
- [x] `npm run build` succeeds for both `en` and `zh` (verified locally, both locales, no errors or warnings)
- [x] Chinese translation updated if English docs changed (or noted why not, see note above)
- [x] Commits are signed off (`git commit -s`)

[...]

PR #719, opened 2026-08-04
fix(i18n): sync zh protocol documentation with english source (#542)

ApurveKaranwal:
**What type of PR is this?**

/kind documentation

**What this PR does / why we need it**:

Fixes #542 by synchronizing the Chinese `developers/protocol.md` documentation with the current protocol design and resolving diagrams inventory inconsistencies.

Per maintainer guidance from @mesutoezdil, older frozen releases (`v2.6.0`, `v2.7.0`, `v2.8.0`) are left intact, while updates target `docs/` + ZH `current/` and are synchronized to `v2.9.0`.

### Summary of Changes

**Chinese Protocol Documentation (`i18n/zh/` current & v2.9.0):**
- Updated device registration to the current node annotation format (`hami.io/node-handshake-{device-type}` and `hami.io/node-{device-type}-register`) with `protocol-register.png`.
- Retained the active "Task Dispatch & Scheduling Decisions" (`任务分发与调度决策`) section documenting `bind-time`, `vgpu-devices-allocated`, and `vgpu-devices-to-allocate` (with `task-dispatch.png`).
- Clarified that `hami.io/vgpu-devices-to-allocate` serializes to `;` (empty list) when complete.
- Corrected the Kubernetes call path: `kube-scheduler` binds the Pod, and `kubelet` calls device-plugin's `Allocate` during container creation.
- Avoided hardcoding annotation counts to account for `bind-phase` and `vgpu-node`.

**Diagrams Inventory (`diagrams-inventory.md` across EN/ZH current & v2.9.0):**
- Updated protocol diagram mappings to record `protocol-register.png` for EN + ZH and `task-dispatch.png` for ZH (EN pending #811).
- Synchronized table row ordering so active `task-dispatch.png` precedes historical-only `device-registration.png`.
- Standardized `Huawei Ascend NPU` naming across English copies.

**Which issue(s) this PR fixes**:

Fixes #542

**Checklist**:
- [x] `npm run lint` and `npm run format:check` pass
- [x] `npm run test` passes (73/73)
- [x] `npm run build` succeeds
- [x] Chinese translation updated if English docs changed (or noted why not)
- [x] Commits are signed off (`git commit -s`)

PR #779, opened 2026-08-15
docs: add GPU node registration troubleshooting guide

ipsitapp8:
## Summary

Adds a troubleshooting guide for GPU nodes that do not register with HAMi.

The guide follows the three registration channels from the device plugin to kubelet, the node annotation, and the scheduler cache. It covers common causes such as node selectors, NVML and runtime failures, kubelet socket paths, RBAC, stale annotations, and scheduler leadership.

The protocol documentation is updated in the same PR because the guide depends on the registration details. In particular, it corrects the handshake timeout from five minutes to 60 seconds, documents the 15-second scheduler loop, and explains the JSON encoding used by NVIDIA. Keeping these changes together avoids having the two pages describe different behavior. The 60 second and 15 second figures are verified directly against `pkg/device/devices.go` (`CheckHealth`) and `pkg/scheduler/scheduler.go` at the `v2.9.0` tag, and match current master too, so this isn't a version-specific correction, the "5 minutes" text was simply wrong at v2.9.0 already.

The guide is included in the current docs and the v2.9.0 version, with English and Chinese pages.

## Validation

- `npx markdownlint` passed for all changed Markdown files.
- `git diff --check` passed.
- Full Docusaurus build passed for English and Chinese.

The build still reports the existing duplicate routes for `/landing/coscup-2026` and `/landing/opensource-summit-korea`.

N/A

**Checklist**:

- [x] Chinese translation updated if English docs changed

**AI assistance disclosure**: This guide and description were drafted with Claude Code, then verified against the HAMi v2.9.0 source and the actual annotation/log behavior myself before opening the PR.

### DIFF
--- a/docs/developers/protocol.md
+++ b/docs/developers/protocol.md
@@ -6,7 +6,7 @@ title: Protocol design
 
 <img src="/img/docs/common/developers/protocol/protocol-register.png" width="600px" alt="HAMi device registration protocol diagram showing node annotation process" />
 
-HAMi needs to know the spec of each AI device in the cluster to schedule properly. During device registration, device-plugin needs to keep patching the spec of each device into node annotations every 30 seconds, in the format of the following:
+HAMi needs to know the spec of each AI device in the cluster to schedule properly. During device registration, device-plugin rescans its devices every 30 seconds and patches the spec of each device into node annotations, in the format of the following:
 
 ```text
 hami.io/node-handshake-\{device-type\}: Reported_\{device_node_current_timestamp\}
@@ -18,6 +18,14 @@ The definition of each device is in the following format:
 ```text
 \{Device UUID\},\{device split count\},\{device memory limit\},\{device core limit\},\{device type\},\{device numa\},\{healthy\}
 ```
+
+:::note Encoding differs by vendor
+
+The colon-separated form above is the legacy encoding, still used by device types decoded with `DecodeNodeDevices` (for example DCU and Iluvatar). NVIDIA and several other device types encode the same fields as a **JSON array** instead, one object per device, decoded with `UnMarshalNodeDevices`. Fields are serialized with `omitempty`, so zero and `false` values are absent rather than written out. See [GPU Virtualization](../core-concepts/gpu-virtualization.md) for a JSON example.
+
+The annotation key is also not fully uniform: NVIDIA uses `hami.io/node-nvidia-register` and `hami.io/node-handshake` (no device-type suffix), while Kunlun uses `hami.io/node-register-xpu`.
+
+:::
 
 An example is shown below:
 
@@ -31,10 +39,18 @@ hami.io/node-nvidia-register: GPU-00552014-5c87-89ac-b1a6-7b53aa24b0ec,10,32768,
 
 In this example, this node has two different AI devices, 2 NVIDIA-V100 GPUs, and 2 Cambricon 370-X4 MLUs
 
-A device node may become unavailable due to hardware or network failure. If a node hasn't registered in the last 5 minutes, the scheduler marks it as 'unavailable'.
-
-Since system clock on scheduler node and 'device' node may not align properly, scheduler node will patch the following device node annotations every 30s
+A device node may become unavailable due to hardware or network failure. Since the system clock on the scheduler node and on the device node may not align properly, the scheduler owns the timestamp: whenever the handshake annotation is absent or does not contain `Requesting`, the scheduler stamps it with its own clock.
 
 ```text
 hami.io/node-handshake-\{device-type\}: Requesting_{scheduler_node_current_timestamp}
 ```
+
+The scheduler's registration loop runs every 15 seconds, and also on node events and on leader election changes. Only the elected leader performs registration.
+
+A handshake is treated as expired once its timestamp is more than **60 seconds** old. Expiry alone does not remove the node. The scheduler additionally requires the node's allocatable device count to have dropped to zero before it runs node cleanup, which removes the node's devices from the scheduler cache and deletes the handshake annotation. A node whose device-plugin is still reporting to kubelet therefore stays available even with an expired handshake.
+
+:::note The NVIDIA device-plugin does not write the handshake
+
+The `Reported_` side of this protocol is written by device-plugins that implement it. The in-tree NVIDIA device-plugin does not, so on an NVIDIA node the annotation normally stays at `Requesting_<timestamp>` and ages past 60 seconds. That is the expected steady state, not a fault: the effective liveness signal for NVIDIA is the allocatable device count. See [GPU Nodes Not Registering](../troubleshooting/node-registration.md).
+
+:::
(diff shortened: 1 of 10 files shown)


ipsitapp8:
@archlitchi heyy! can you review the pr for me?

versioned_docs/version-v2.9.0/developers/protocol.md
```diff
@@ -31,10 +39,18 @@ hami.io/node-nvidia-register: GPU-00552014-5c87-89ac-b1a6-7b53aa24b0ec,10,32768,
 
 In this example, this node has two different AI devices, 2 Nvidia-V100 GPUs, and 2 Cambricon 370-X4 MLUs
 
-A device node may become unavailable due to hardware or network failure. If a node hasn't registered in the last 5 minutes, the scheduler marks it as 'unavailable'.
-
-Since system clock on scheduler node and 'device' node may not align properly, scheduler node will patch the following device node annotations every 30s
+A device node may become unavailable due to hardware or network failure. Since the system clock on the scheduler node and on the device node may not align properly, the scheduler owns the timestamp: whenever the handshake annotation is absent or does not contain `Requesting`, the scheduler stamps it with its own clock.
```
Creativeklvn:
```suggestion
A device node may become unavailable due to hardware or network failure. Since the system clock on the scheduler node and on the device node may not align properly, the scheduler owns the timestamp. Whenever the handshake annotation is absent or does not contain `Requesting`, the scheduler stamps it with its own clock.
```

Using a full stop here makes the sentence easier to read.

versioned_docs/version-v2.9.0/developers/protocol.md
```diff
@@ -31,10 +39,18 @@ hami.io/node-nvidia-register: GPU-00552014-5c87-89ac-b1a6-7b53aa24b0ec,10,32768,
...
 
 ```text
 hami.io/node-handshake-\{device-type\}: Requesting_{scheduler_node_current_timestamp}
 ```
+
+The scheduler's registration loop runs every 15 seconds, and also on node events and on leader election changes. Only the elected leader performs registration.
+
+A handshake is treated as expired once its timestamp is more than **60 seconds** old. Expiry alone does not remove the node. The scheduler additionally requires the node's allocatable device count to have dropped to zero before it runs node cleanup, which removes the node's devices from the scheduler cache and deletes the handshake annotation. A node whose device-plugin is still reporting to kubelet therefore stays available even with an expired handshake.
+
+:::note The NVIDIA device-plugin does not write the handshake
```
Creativeklvn: I think it would be clearer to change this to:'.

```suggestion
:::note The NVIDIA device-plugin does not write the `Reported_` handshake
```

Saying it does not write the handshake could make it sound like NVIDIA does not participate in the handshake at all.

docs/developers/protocol.md
```diff
@@ -31,10 +39,18 @@ hami.io/node-nvidia-register: GPU-00552014-5c87-89ac-b1a6-7b53aa24b0ec,10,32768,
...
 
 ```text
 hami.io/node-handshake-\{device-type\}: Requesting_{scheduler_node_current_timestamp}
 ```
+
+The scheduler's registration loop runs every 15 seconds, and also on node events and on leader election changes. Only the elected leader performs registration.
+
+A handshake is treated as expired once its timestamp is more than **60 seconds** old. Expiry alone does not remove the node. The scheduler additionally requires the node's allocatable device count to have dropped to zero before it runs node cleanup, which removes the node's devices from the scheduler cache and deletes the handshake annotation. A node whose device-plugin is still reporting to kubelet therefore stays available even with an expired handshake.
+
+:::note The NVIDIA device-plugin does not write the handshake
```
Creativeklvn:
```suggestion
:::note The NVIDIA device-plugin does not write the `Reported_` handshake
```

docs/developers/protocol.md
```diff
@@ -19,6 +19,14 @@ The definition of each device is in the following format:
 \{Device UUID\},\{device split count\},\{device memory limit\},\{device core limit\},\{device type\},\{device numa\},\{healthy\}
 ```
 
+:::note Encoding differs by vendor
+
+The colon-separated form above is the legacy encoding, still used by device types decoded with `DecodeNodeDevices` (for example DCU and Iluvatar). NVIDIA and several other device types encode the same fields as a **JSON array** instead, one object per device, decoded with `UnMarshalNodeDevices`. Fields are serialized with `omitempty`, so zero and `false` values are absent rather than written out. See [GPU Virtualization](../core-concepts/gpu-virtualization.md) for a JSON example.
+
+The annotation key is also not fully uniform: NVIDIA uses `hami.io/node-nvidia-register` and `hami.io/node-handshake` (no device-type suffix), while Kunlun uses `hami.io/node-register-xpu`.
```
mesutoezdil, line 26: this note says the nvidia handshake key is unsuffixed and never reaches reported, but the example a few lines below still shows hami.io/node-handshake-nvidia: Reported. after this pr the page contradicts itself, fix the example too.

docs/troubleshooting/node-registration.md
```diff
@@ -0,0 +1,384 @@
+---
+title: GPU Nodes Not Registering
```
mesutoezdil, line 2: which hami version were the 60 second expiry and the 15 second loop verified against? #719 and the current page both say 5 minutes, and this pr backports the correction into the frozen v2.9.0 snapshot. if the numbers changed after v2.9.0 the snapshot must keep the old ones.

mesutoezdil (commented):
very strong guide, the three channel model and the omitempty warning are exactly what operators need. blockers and coordination: dco fails, the two suggestion commits have no sign off, squash and sign. this rewrites protocol.md in every locale and version, which collides head on with #719, and the sidebar block collides with #741 and #767, agree on an order. the pr body has no template, no issue link and no ai note.

docs/developers/protocol.md
ipsitapp8: Fixed the example. Changed hami.io/node-handshake-nvidia: Reported ... to hami.io/node-handshake: Requesting_..., matching both the unsuffixed-key and does-not-write-Reported_ notes. Applied the same fix to the versioned_docs/version-v2.9.0 snapshot, it had the identical contradiction.

docs/developers/protocol.md
ipsitapp8: This suggestion landed on the versioned_docs/version-v2.9.0 copy via the commit-suggestion button but never got mirrored to this current page. Applied it here now.

docs/troubleshooting/node-registration.md
ipsitapp8: Checked the HAMi repo history. The 60 second expiry has been time.Second * 60 in pkg/device/devices.go since commit f95852b (2025-09-01), unchanged since. The 15 second registration loop has been time.NewTicker(time.Second * 15) in pkg/scheduler/scheduler.go since commit 4b3876e (2024-03-20), unchanged since. The v2.9.0 tag was cut 2026-05-19, well after both. So the numbers were already 60s/15s at v2.9.0, the 5 minutes in the current page and #719 was just wrong from the start, not an old value that changed later. Backporting the correction into the v2.9.0 snapshot is accurate.

mesutoezdil:
Closing this for now: several open PRs add new troubleshooting pages and sidebar entries that conflict with each other and with the reorganization tracked in #689.

PR #785, opened 2026-08-17
docs(sidebar): add missing contributor guide entries

ipsitapp8:
**What type of PR is this?**

/kind documentation

Registers four existing documentation files under `docs/contributor/` into `sidebars.js`:

- `contributor/cherry-picks` (How to cherry-pick PRs)
- `contributor/lifted` (How to manage lifted code)
- `contributor/adopters` (HAMi Adopters list and submission guide)
- `contributor/contributors` (Contributors list)

These documents were present in the repository but not listed in `sidebars.js`, making them unreachable through the site's sidebar navigation.

N/A

**Checklist**:

- [x] Chinese translation updated if English docs changed (N/A: only navigation structure updated in `sidebars.js`)


rootsongjc:
The reason these documents are not included in the sidebar is that I don't think all of them need to be kept. Some of them contain outdated or incorrect information, such as the roadmap and contributor list, which are no longer necessary to maintain on the website.

There are also some basic Git operation instructions that I don't think need to be part of the Contributor Guide. Instead of keeping these documents as they are, I think we should revisit and reorganize the structure and content of the entire `Contributor` directory, remove outdated materials, and make sure the remaining documentation is useful and maintainable for contributors.

ipsitapp8:
@rootsongjc  I only wired existing docs into the sidebar without checking whether they still hold up. Could you point me at which ones you'd cut (you mentioned roadmap and contributor list as outdated, plus the basic git instructions as out of scope for a Contributor Guide)? I'll narrow this PR to just the docs that are still accurate and worth surfacing, and leave the bigger Contributor directory reorganization as a separate follow-up rather than trying to fold it into this one.

sidebars.js
```diff
@@ -412,10 +412,14 @@ module.exports = {
         "contributor/contributing",
         "contributor/contribute-docs",
         "contributor/github-workflow",
+        "contributor/cherry-picks",
```
mesutoezdil, line 415: the same four files exist in the v2.9.0 snapshot and its sidebar has the same gap. v2.9.0 is the default version visitors land on, so without updating versioned_sidebars/version-v2.9.0-sidebars.json the pages stay unreachable for most readers. also a nit: lifted sits between roadmap and e2e-testing, grouping it next to cherry-picks reads better.

mesutoezdil (commented):
verified all four files exist in en and zh, none are marked unlisted, ci is fully green. one gap inline.

sidebars.js
ipsitapp8: Added the same four entries to versioned_sidebars/version-v2.9.0-sidebars.json, and moved lifted next to cherry-picks in both sidebars. Verified with a full en+zh build.

ipsitapp8:
@mesutoezdil can you check the pr now if you have time?

mesutoezdil:
Closing this for now: the reorganization tracked in #689 rewrites the sidebar, so these entries would need to be redone right after it lands.

PR #819, opened 2026-08-24
fix(tutorials): match HAMi time.DateTime layout in handshake annotation

WaliTheOrchestrator:
**What type of PR is this?**

/kind documentation

Lab 2 tells the reader to write the hami.io/node-handshake annotation with a dot-separated timestamp:

```bash
hami.io/node-handshake="Requesting_$(date '+%Y.%m.%d %H:%M:%S')"
```

But HAMi writes and parses that value with Go time.DateTime, which is "2006-01-02 15:04:05":

- write: https://github.com/Project-HAMi/HAMi/blob/4707fb0/pkg/device/devices.go#L601
- parse: https://github.com/Project-HAMi/HAMi/blob/4707fb0/pkg/device/devices.go#L583

Go requires literal separators in a layout to match exactly, so a dotted value never parses. It requires hyphens. The lab teaches a format HAMi itself does not produce.

The dotted format was correct once. HAMi wrote "2006.01.02 15:04:05" up to v2.4.1. Project-HAMi/HAMi#741 replaced it with time.DateTime for v2.5.0 in Feb 2025. This lab was written later and carried the old format forward.

**What this does not fix**

This does not fix a visible failure. When the parse fails CheckHealth returns (true, false) which is the same value it returns for a valid fresh handshake and the error is never logged. The lab completes either way.

I transcribed CheckHealth into a small Go program and ran it on Go 1.23.12. It returned:

```plaintext
A. Hyphens, seconds old                  ->  (true,  false)  HEALTHY  parsed, fresh
B. Hyphens, 2 hours old, no allocatable  ->  (false, false)  DEAD     parsed, expired
C. Dots,    seconds old                  ->  (true,  false)  HEALTHY  PARSE FAILED
D. Dots,    2 hours old, no allocatable  ->  (true,  false)  HEALTHY  PARSE FAILED
```

B and D differ only by the separator. A and C return identical values, one after a successful parse and one after a failure.

**Why it still matters**

Nothing breaks in the lab, but the format the lab teaches would matter on a real cluster. Because a dotted value never parses, HAMi can never mark that node unhealthy. A node whose device plugin has died keeps being reported as alive, and nothing is logged, so nobody finds out. It is a dead node that still looks healthy.

Go is strict about separators for a reason, and the dot and the hyphen are not interchangeable here. This is also the first lab a new user runs, so it is where people learn what a HAMi timestamp is supposed to look like.

**Scope**

Two files: the English lab and its Chinese mirror. These are the only two places in the repository where a reader is instructed to run a command that produces a dotted timestamp.

docs/developers/protocol.md and several versioned_docs snapshots also show dotted samples. I left those out to keep this to one change, and can open a separate issue if that would help.

The Chinese page states that the payload matches what the v2.9.0 device plugin writes on a real GPU node. That was not true for the handshake value before this
change.

**Which issues this PR fixes**:

None filed.

**Checklist**:

- [x] npm run lint and npm run format:check pass
- [x] npm run build succeeds for both en and zh
- [x] Chinese translation updated if English docs changed
- [x] Commits are signed off (git commit -s)

## Summary by CodeRabbit

* **Documentation**
  * Updated the local fake GPU tutorial to use the standardized `YYYY-MM-DD` format for HAMi WebUI node handshake timestamps.
  * Applied the same formatting correction to the Chinese tutorial translation.


windsonsea (commented):
/approve

tutorials/labs/local-fake-gpu.md
```diff
@@ -724,7 +724,7 @@ In a real environment, the HAMi device-plugin automatically writes the `hami.io/
 ```bash
 kubectl annotate node ${NODE_NAME} \
   hami.io/node-nvidia-register='[{"id":"GPU-3cef3724-8228-5a66-b391-b0901788f5d0","count":10,"devmem":11441,"devcore":100,"type":"NVIDIA-Tesla-K80","mode":"hami-core","health":true},{"id":"GPU-5127182e-f297-5a25-bb44-0444c3be540c","index":1,"count":10,"devmem":11441,"devcore":100,"type":"NVIDIA-Tesla-K80","mode":"hami-core","health":true}]' \
-  hami.io/node-handshake="Requesting_$(date '+%Y.%m.%d %H:%M:%S')"
+  hami.io/node-handshake="Requesting_$(date '+%Y-%m-%d %H:%M:%S')"
```
mesutoezdil, line 727: correction to my earlier note: pkg/device/nvidia/device.go defines HandshakeAnnos as the bare `hami.io/node-handshake`, so this key is right and protocol.md is the page that is wrong. only the date format question stands, and `%Y-%m-%d` matches time.DateTime, so this looks correct.

mesutoezdil, line 727: confirmed the fix: pkg/device/devices.go:600 parses this with time.DateTime, so `%Y-%m-%d %H:%M:%S` is exactly right.

tutorials/labs/local-fake-gpu.md
```diff
@@ -724,7 +724,7 @@ In a real environment, the HAMi device-plugin automatically writes the `hami.io/
 ```bash
 kubectl annotate node ${NODE_NAME} \
   hami.io/node-nvidia-register='[{"id":"GPU-3cef3724-8228-5a66-b391-b0901788f5d0","count":10,"devmem":11441,"devcore":100,"type":"NVIDIA-Tesla-K80","mode":"hami-core","health":true},{"id":"GPU-5127182e-f297-5a25-bb44-0444c3be540c","index":1,"count":10,"devmem":11441,"devcore":100,"type":"NVIDIA-Tesla-K80","mode":"hami-core","health":true}]' \
```
mesutoezdil, line 726: devices.go:618 has the scheduler stamp Requesting_ itself when this annotation is missing. does the lab still need to set it by hand, or does that only save one reconcile cycle?

mesutoezdil:
/lgtm

the change is right. `pkg/device/devices.go:600` parses that timestamp with `time.ParseInLocation(time.DateTime, ...)`, and Go`s `time.DateTime` is `2006-01-02 15:04:05`, so `%Y-%m-%d %H:%M:%S` is exactly the layout it expects. the old dotted form would not parse.

i resolved my own two threads, they were confirmations rather than questions. i left the third one open, but it is not blocking: it asks whether the lab needs to write the handshake annotation by hand at all, since `devices.go:618` has the scheduler stamp it when it is missing. that is about a step this pr does not touch, so it can be a follow up.

rootsongjc (approved):
/LGTM

PR #828, opened 2026-08-28
docs: clarify AWS Marketplace version

### LINKED ISSUE #827: Update AWS Installation Version Reference
**Page URL**:
https://project-hami.io/docs/installation/aws-installation

**What is wrong**:
The AWS installation documentation hard-codes the HAMi version 2.6.1-community. If a newer version is released on the AWS Marketplace in the future, the installation instructions could become outdated.

**What it should say instead**:
Update the documentation to clarify that `2.6.1-community` is the version used in the guide and provide a link to the AWS Marketplace listing for users who want to check the latest version.

Creativeklvn:
**What type of PR is this?**
/kind documentation

Clarifies the AWS Marketplace version used in the installation instructions.

- Specifies that the installation uses version `2.6.1-community`.
- Directs users to check the AWS Marketplace for the latest available version.

Fixes #827

**Checklist**:

- [ ] Chinese translation updated if English docs changed (or noted why not)

## Summary by CodeRabbit

- **Documentation**
  - Updated the AWS installation guide to note availability through the AWS Marketplace.
  - Added installation options using Helm or the AWS add-on.
  - Documented the `2.6.1-community` version and instructed users to check the AWS Marketplace listing for the latest version.

### DIFF
--- a/docs/installation/aws-installation.md
+++ b/docs/installation/aws-installation.md
@@ -3,7 +3,7 @@ title: HAMi on AWS
 translated: true
 ---
 
-HAMi is now available on the AWS Marketplace, and you can quickly install it via Helm or AWS add-on.
+HAMi is now available on the AWS Marketplace and can be installed via Helm or the AWS add-on.
 
 ## Prerequisites
 
@@ -13,6 +13,8 @@ Before installation, please ensure you have:
 - Created a Kubernetes cluster
 
 ## Install with Helm
+
+The installation below uses version `2.6.1-community`. To install a newer version, replace `2.6.1-community` in the Helm command with the latest version available on the [AWS Marketplace listing](https://aws.amazon.com/marketplace/pp/prodview-x237accxbpuwa).
 
 You can use the following commands to pull HAMi’s Helm chart and install it:
 


docs/installation/aws-installation.md
```diff
@@ -3,7 +3,7 @@ title: HAMi on AWS
 translated: true
 ---
 
-HAMi is now available on the AWS Marketplace, and you can quickly install it via Helm or AWS add-on.
+HAMi is now available on the AWS Marketplace and can be installed via Helm or the AWS add-on. The installation below uses version 2.6.1-community; for the latest available version, check the [AWS Marketplace listing](https://aws.amazon.com/marketplace/pp/prodview-x237accxbpuwa).
```
mesutoezdil: the helm command below still pins `--version 2.6.1-community`. a reader who checks the marketplace for a newer version has no way to use it here. should the command say that too?

docs/installation/aws-installation.md
```diff
@@ -3,7 +3,7 @@ title: HAMi on AWS
...
 
 ## Prerequisites
```
mesutoezdil, line 8: this page has translated: true and a zh mirror under i18n/zh, which this pr does not touch. should the version note go there too?

docs/installation/aws-installation.md
mesutoezdil: one more here: the Install with AWS Add-on section below carries no version at all, so the installation below wording covers a path this does not apply to. should the note sit under the helm heading instead?

docs/installation/aws-installation.md
Creativeklvn: I don’t understand Chinese, so I wouldn’t be able to update the `zh` translation accurately. Would it be better to leave the Chinese mirror unchanged and have the translation updated separately?

Creativeklvn:
I have moved the version note under the Install with Helm heading since it only applies to the Helm installation and removed it from the introductory paragraph.

docs/installation/aws-installation.md
```diff
@@ -14,6 +14,8 @@ Before installation, please ensure you have:
 
 ## Install with Helm
 
+The installation below uses version `2.6.1-community`. To install a newer version, replace `2.6.1-community` in the Helm command with the latest version available on the [AWS Marketplace listing](https://aws.amazon.com/marketplace/pp/prodview-x237accxbpuwa).
```
togettoyou, line 17: Non-blocking: Could we add this to the Chinese page as well, either in this PR or a follow-up PR?

Suggested translation:

```
以下安装步骤使用的是 `2.6.1-community` 版本。如需安装更新版本，请将 Helm 命令中的 `2.6.1-community` 替换为 [AWS Marketplace 页面](https://aws.amazon.com/marketplace/pp/prodview-x237accxbpuwa) 上提供的最新版本。
```

Here’s another small suggestion: would the version information be easier to notice if you wrapped it in a `:::note` block?

Image

Image

togettoyou (changes requested):

docs/installation/aws-installation.md
togettoyou: @rootsongjc Perhaps we could merge this first, and then I could submit a Chinese version?

Also, do you think it would be better to add `:::note` blocks?

PR #752, opened 2026-08-10
fix: center and properly size event banners regardless of aspect ratio (#746)

Akashstztt:
Related to #746

**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

The event banner on `/landing/kcd-vietnam` was being cropped because `.banner` used `object-fit: cover`, which is designed for wide/landscape images. This event's banner is square, so key visual content (title, speaker names) was getting cut off.

Fixed by wrapping the banner image in a flex-centered container and using `max-width`/`max-height` with `object-fit: contain` and `width/height: auto`, so the image sizes itself based on its actual aspect ratio instead of being forced to stretch full-width. This displays square banners fully and centered, while landscape banners (e.g. `/landing/kubecon-japan`) still fill the width naturally.

**Which issue(s) this PR fixes**:

Fixes #746

**Checklist**:

- [ ] `npm run lint` and `npm run format:check` pass — *note: both report pre-existing repo-wide issues unrelated to this PR (a Windows markdownlint CLI quirk, and 482 files with existing Prettier formatting drift). Neither of the two files changed in this PR is affected.*
- [ ] `npm run build` succeeds for both `en` and `zh` — *not verified locally due to a Windows-specific `NODE_OPTIONS` shell syntax issue in this environment; CI will validate this on Linux.*
- [x] Chinese translation updated if English docs changed (or noted why not) — N/A, CSS/layout-only change, no text content changed
- [x] Commits are signed off (`git commit -s`)

**Testing**:

- Verified `/landing/kcd-vietnam` banner (square image) now displays fully, with no cropping or letterboxing, sized to its natural aspect ratio.
- Verified `/landing/kubecon-japan` banner (landscape image) still displays correctly at full width, with no regression.
- Confirmed `EventLanding.module.css` is only imported by `EventLanding.js`, which is only used by these two event pages — no other pages affected by this change.

Screenshot 2026-08-22 173035
Screenshot 2026-08-22 172958
Screenshot 2026-08-22 173007

## Summary by CodeRabbit

* **Style**
  * Improved event banner presentation with a centered, contained layout.
  * Added consistent dimensions, rounded corners, border styling, spacing, and overflow handling.
  * Ensured banner images fill the display area without distortion.

PR #834, opened 2026-09-01
fix(event-landing): enforce 16:9 aspect ratio on banner

### LINKED ISSUE #746: [UI] Fix KCD Vietnam event banner image sizing and cropping
## Description

The banner image on `/landing/kcd-vietnam` gets cropped due to `object-fit: cover` not suiting its square aspect ratio. Important visual content is lost because the current CSS is designed for wide/landscape banners, but this event's banner image is square.

## Steps to Reproduce

1. Visit https://project-hami.io/landing/kcd-vietnam and https://project-hami.io/landing/kubecon-japan
2. Observe the event banner at the top of the page

## Expected Behavior

The banner image should display without unwanted cropping, regardless of its aspect ratio. Possible approaches:
- Use `object-fit: contain` instead of `cover` for the `.banner` class
- Adjust the container's height/aspect-ratio to better match square images
- Add per-event configuration for banner display mode

## Screenshot

(attach the screenshot of the cropped banner here — drag it into this box)

## Related

Noticed while working on #742. Tracked separately since it's an unrelated visual/CSS issue.

Image

fishman:
Replace the fixed max-height with an aspect-ratio so the banner scales
responsively, and center the image with object-position.

**What type of PR is this?**

/kind cleanup

Some banners have an odd aspect-ratio. We should enforce 16:9 as default.

Also requires the same aspect-ratio from the KubeCon China banner. @rootsongjc

Fixes #746

Also supercedes #752

**Checklist**:

### DIFF
--- a/src/components/EventLanding.module.css
+++ b/src/components/EventLanding.module.css
@@ -8,9 +8,11 @@
 }
 
 .banner {
+  display: block;
   width: 100%;
-  max-height: 480px;
+  aspect-ratio: 16 / 9;
   object-fit: cover;
+  object-position: center;
   border-radius: var(--hami-radius-md);
   border: 1px solid var(--hami-color-border);
   margin-bottom: var(--hami-space-32);


src/components/EventLanding.module.css
```diff
@@ -8,9 +8,11 @@
 }
 
 .banner {
+  display: block;
   width: 100%;
-  max-height: 480px;
+  aspect-ratio: 16 / 9;
```
mesutoezdil, line 13: this drops the 480px cap, so on a 1140px container the banner is now about 640px tall. in #746 the ask was that banners already look too large on desktop. does this make that worse?

mesutoezdil, line 13: the kcd vietnam banner is 1920x1920 and the kubecon jp one is 1920x1080. so cover cuts 44 percent off the square one, which is the banner #746 is about, and leaves the other untouched. is contain plus a background closer to what you want?

src/components/EventLanding.module.css
```diff
@@ -8,9 +8,11 @@
...
   object-fit: cover;
+  object-position: center;
```
mesutoezdil, line 15: object-position already defaults to center, so this line changes nothing. keep it as documentation, or drop it?

src/components/EventLanding.module.css
fishman: No

fishman: Everything cuts a part off, the Vietnam banner is 4:3 but allows for 16:9 to keep all the important content.

src/components/EventLanding.module.css
fishman: Not sure, will double check but if true, probably drop it. The other pr that set a fixed width seemed to have caused it to be left aligned, but maybe I saw it wrong.

src/components/EventLanding.module.css
fishman: The bigger issue is the Kubecon China banner, but that's a normal format. Almost all conference banners are 16/9

mesutoezdil: i measured the file on master and it is 1920x1920, so 1:1 rather than 4:3. at 16:9 that drops 44 percent of the height instead of 25. still fine for the important content?

mesutoezdil:

PR #869, opened 2026-09-15
fix(ui): keep sidebar aligned with fixed header

togettoyou:
**What type of PR is this?**

/kind cleanup

This PR fixes the Docs and Tutorials sidebars moving upward when the page starts scrolling while the announcement bar is visible. It adjusts the sticky sidebar offset and available height to account for the fixed announcement bar, keeping the sidebar aligned below the header. When the announcement bar is dismissed, the default sidebar layout is restored automatically.

**Checklist**:

- [X] `npm run lint` and `npm run format:check` pass
- [X] `npm run build` succeeds for both `en` and `zh`
- [X] Chinese translation updated if English docs changed (or noted why not)
- [X] Commits are signed off (`git commit -s`)

**AI assistance disclosure**

The fix for the sidebar layout in this PR was completed with the assistance of Codex. I reviewed the CSS changes and verified the sidebar's behavior locally.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved documentation navigation layout on larger screens so the sidebar clears the announcement bar and remains within the visible page area.


src/css/custom.css
```diff
@@ -305,6 +305,16 @@ body,
   .theme-announcement-bar ~ .main-wrapper {
     padding-top: calc(var(--ifm-navbar-height) + var(--docusaurus-announcement-bar-height, 0px));
   }
+
+  .theme-announcement-bar ~ .main-wrapper .theme-doc-sidebar-container > div {
+    top: var(--docusaurus-announcement-bar-height, var(--hami-announcement-offset, 30px));
+    height: calc(
+      100vh - var(--docusaurus-announcement-bar-height, var(--hami-announcement-offset, 30px))
+    );
+    max-height: calc(
```
mesutoezdil, line 314: since dismissing the bar only updates this sibling selector live, could the toc's offset above (gated on the static `data-announcement-bar-initially-dismissed` attribute) end up stuck with the extra gap until the next page load, out of sync with this sidebar?

togettoyou: You are right. this issue was already there. I'll fix it.

mesutoezdil:
pls check the reviews of coderabbit

src/css/custom.css
```diff
@@ -305,6 +305,16 @@ body,
...
+      100vh - var(--docusaurus-announcement-bar-height, var(--hami-announcement-offset, 30px))
```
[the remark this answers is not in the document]
togettoyou: Docusaurus already adds `padding-top: var(--ifm-navbar-height)` to the inner sidebar, so this change would count the navbar twice and leave an extra 60px gap. The current values keep the menu aligned below the navbar.

src/css/custom.css
togettoyou: Done.

togettoyou:
> pls check the reviews of coderabbit

done.

mesutoezdil (approved):
LGTM

